PR #275, opened 2014-12-28
Fix : Can't handle CJK multibyte request in AJP parameter value

nameislocus:
### Problem :

In HTTP protocol hander can handle multibyte character set. My case use EUC-KR character set.  Tested in Wildfly 8.2 GA.
```xml
            <server name="default-server">
                <ajp-listener name="ajp" socket-binding="ajp" url-charset="EUC-KR"/>
                <http-listener name="default" socket-binding="http" url-charset="EUC-KR"/>
```
http://localhost:8080/test/test.jsp?param=한글

But, When I use AJP protocol can't handle EUC-KR character set.
http://localhost/test/test.jsp?param=한글

Query param's value are not '한글', it  prints '??'.
Similar cases are in https://developer.jboss.org/thread/242677?start=0&tstart=0

### Reason :

I found that it cause of handle byte as a char in AbstractAjpParser. 
Multibyte(CJK ; Chinese, Japanese, Korean) character set are use most significant bit. So, when byte are casted to char, multibytes informations are eliminated.

Thanks for your great **'undertow'**.

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/ajp/AbstractAjpParseState.java
+++ b/core/src/main/java/io/undertow/server/protocol/ajp/AbstractAjpParseState.java
@@ -18,6 +18,8 @@
 
 package io.undertow.server.protocol.ajp;
 
+import java.util.ArrayList;
+
 /**
  * Abstract AJP parse state. Stores state common to both request and response parsers
  *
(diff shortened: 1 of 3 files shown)


stuartwdouglas:
Thanks for the PR, however I have fixed it a different way that should give better performance:
https://github.com/undertow-io/undertow/commit/a567f4ed7f7e23c317182914ee384d82320c4da2

Can you test this out and make sure it works for you?

nameislocus:
It works very well. 
Thanks for your fix.

PR #283, opened 2015-02-10
I met  char  garbled for  chinese 

KBCode:
I met Char garbled for chinese, the charset is utf-8. 
I read source find position code.
io.undertow.server.handlers.form.FormEncodedDataDefinition.FormEncodedDataParser.doParse(final StreamSourceChannel channel)

//some byte is minus  
builder.append((char) n);
//the garbled char by save 
data.add(name, builder.toString());

//I get the string is garbled
request.getParameter("name");

I how get correct for String?

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -25,12 +25,12 @@
     <parent>
         <groupId>io.undertow</groupId>
         <artifactId>undertow-parent</artifactId>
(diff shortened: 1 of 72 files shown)


ctomc:
what is this pull request suppose to be? I don't 'see any of your commits in it.
If you just want to report issue, than you should use jira for that

PR #332, opened 2015-09-06
UNDERTOW-488 Allow setting the same cookie more than once

aldaris:
This patch should resolve UNDERTOW-488 by ensuring that response cookies can be set multiple times with the same name=value mapping.


undertow-pull-request:

ctomc:
This is ok to test

PR #346, opened 2015-12-18
1.1.x

fedorpatlin:
This prevents error message java.lang.IllegalAccessError in clustered installation. See http://stackoverflow.com/questions/34304940/clustered-wildfly-throws-java-lang-illegalaccesserror


stuartwdouglas:
I am going to fix this in a slightly different way. Your patch is racey, because the field is shared it is possible another thread would have set it back to not being accessible and you will still get the exception.

PR #356, opened 2016-02-02
[UNDERTOW-626] Using modcluster filter with local content creates loop

AdrianFarmadin:
Possible solution to https://issues.jboss.org/browse/UNDERTOW-626

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 Undertow
 ========
 
+Fork changes:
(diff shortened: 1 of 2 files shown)


undertow-pull-request:

stuartwdouglas:
This sort of fix belongs in Wildfly, rather than Undertow, as it deals with Wildfly specific concepts.

PR #367, opened 2016-03-03
Make sure that getOutputStream() has not been called when using getWriter()

gdrouet:
##When Undertow writes an error, it always calls getWriter(). The method can throw an exception if getOutputStream() has been previously called in the response's lifecycle.

This PR proposes to use getOutputStream() in that case to avoid any error thrown by the servlet container.


undertow-pull-request:

stuartwdouglas:
I have addressed this in a different way (by resetting the writer state before doing an error dispatch). See https://issues.jboss.org/browse/UNDERTOW-654

gdrouet:
Ok great.

PR #380, opened 2016-04-11
Add null-checks when stopping server

popstr:
Undertow should not throw a NullPointerException if stopping a server
that is not correctly started.


popstr:
Adding this to 1.4.x as well.

undertow-pull-request:

stuartwdouglas:
This was already cherry picked to 1.4.x

PR #382, opened 2016-04-13
[UNDERTOW-688] Make all methods on SessionListener default / no-op so  implementations can choose which ones they are interested in.

darranl:


n1hility:
retest this please

stuartwdouglas:
I had been holding off on this until I decided what to do with JDK8 in 1.4.x, we have decided to require JDK8 so this is ok

PR #390, opened 2016-04-22
[UNDERTOW-548] Adjustment to status code handling to better fit FORM auth with other challenge types.

darranl:


servlet/src/main/java/io/undertow/servlet/handlers/security/ServletFormAuthenticationMechanism.java
```diff
@@ -149,6 +155,37 @@ protected void handleRedirectBack(final HttpServerExchange exchange) {
...
+
+        private int status = OK;
+
+        private FormResponseWrapper(HttpServletResponse response) {
+            super(response);
+        }
+
+        @Override
+        public void setStatus(int sc, String sm) {
+            if (super.getStatus() == OK) {
```
stuartwdouglas: I think this check should happen before applying the wrapper (i.e. if the current status code is 200 don't apply the wrapper, if it is not 200 don't let it be modified). Otherwise if there are multiple calls to setStatus the first one may work and the second will fail.

stuartwdouglas:
I think this is too big a change for 1.3.x, it will go into 1.4.x and master.

PR #408, opened 2016-07-06
Extend PathTemplate (match method) to support wildcard matching in path string

ArekCzarnik:
hallo *,

i extend PathTemplate to support wildcard matching in path string
for example path url like /docs/{docId}/* match /docs/test/my.

i hope that is ok for the first pull request.

Thanks,
Arek


undertow-pull-request:

ctomc:
this is ok to test

stuartwdouglas:
Oops, I did not notice that this was targetting 1.3.x. As this is a new feature this will need to go into 1.4.x and 2.x. 1.4.x should have a final release very shortly.

stuartwdouglas:
I have reverted the 1.3.x change and cherry-picked it only master and 1.4.x

ArekCzarnik:
Thanks for the merge, sorry that i not request to 1.4.x.

PR #409, opened 2016-07-08
[UNDERTOW-766] move trace logging for SSO code in Undertow to use 'SECURITY_LOGGER'

iweiss:
Issue: https://issues.jboss.org/browse/UNDERTOW-766

PR #410, opened 2016-07-08
[JBEAP-5152] move trace logging for SSO code in Undertow to use 'SECURITY_LOGGER'

iweiss:
Issue: https://issue.jboss.org/browse/JBEAP-5152
Upstream issue: https://issues.jboss.org/browse/UNDERTOW-766
Upstream PR: https://github.com/undertow-io/undertow/pull/409


stuartwdouglas:
The problem with this is that TRACE can be quite verbose. If these have their own category it is a lot easier to just see the parts that you are interested in.

iweiss:
I tested this PR modifying logging.properties to just TRACE `logger.io.undertow.request.security.level` and it worked as expected. I think that was the purpose of the issue initially.

stuartwdouglas:
I don't want to do this. Trace logging should be in its own category to allow for more fine grained control.

PR #416, opened 2016-08-17
Add fonts mime types

fermin-silva:
Because they were giving me problems in Chrome when serving static font files


undertow-pull-request:

PR #420, opened 2016-08-23
[UNDERTOW-804] fix incorrect session-avg-alive-time value.

soul2zimate:
https://issues.jboss.org/browse/UNDERTOW-804

In case of only one session, it should not check sessionManager.averageSessionLifetime.get() as it is zero.
Previous count is incorrect as BigDecimal is immutable.
A small adjustment on expiredSessionCount as I count averageSessionLifetime after longestSessionLifetime to reuse variable life.

PR #421, opened 2016-08-23
[UNDERTOW-804] fix incorrect session-avg-alive-time value.

soul2zimate:
https://issues.jboss.org/browse/UNDERTOW-804

upstream PR: https://github.com/undertow-io/undertow/pull/420


stuartwdouglas:
I think this whole approach may need to change, as there are some thread safety issues. In particular expiredSessionCount and averageSessionLifetime are not updated atomically.

stuartwdouglas:
I have changed this so it now has a sync block. I think there may have been issues with rounding errors as well as expired sessions get large.

PR #428, opened 2016-09-14
SPDY Connection: Make synchronized requests

jwendell:
In the SPDY protocol the order of streams is important.
Multi-thread apps that use undertow to interact with SPDY servers
must use a synchronized request in order to guarantee that streams
don't arrive in a messed order in the endpoint.


undertow-pull-request:

PR #436, opened 2016-10-13
Return default charset if Content-Type header is absent.

twz123:
To be par with the Javadoc on `getRequestCharset()` and `getResponseCharset()`, also return `ISO_8859_1` instead of `null` in the case of an absent Content-Type header.


undertow-pull-request:

PR #439, opened 2016-10-21
UNDERTOW-545 - SPNEGO test fixed.

rhatlapa:
* Updated enctypes in krb5.conf to enctypes properly working also on IBM JDK
* Setup environment for krb5.conf location moved before actually starting the server.
* Solved issue when token decoded from the header contains improperly 0 bytes at the end, making the token byte[] corrupted by using Base64.decode.


rhatlapa:
The 0 bytes are caused by decode method not taking into account padding bytes when counting size of the underlying array, resulting in bigger array than needed in some cases.

Base64 has this counting handled, see `java.util.Base64.Decoder#outLength`

PR #440, opened 2016-10-31
Relative paths are not properly resolved when base is a symlink

cakofony:
PathResourceManager will not properly resolve `./test.txt` although even if it does resolve `test.txt` when the base path is a symlink.

Added a test that will reproduce the problem, and the fix in a separate commit. I'm not sure how you guys feel about the new `test-assets` directory that includes a symlink. When symlinks are created in source sets, the symlink is exploded into a copy at compile time.

It may be worthwhile to use something like [jimfs](https://github.com/google/jimfs) for this sort of testing, though it pulls in guava which I don't see in the undertow codebase.


undertow-pull-request:

PR #456, opened 2016-12-11
enhanced routing handler to support wildcards of pathtemplates

krampenschiesser:
I have beeen using undertow routing manager and discovered that wildcard support was kinda off.
**/bla/{blubb}/*** is supported where as 
**/bla/*** is not supported.
So I digged into it and discovered that the _PathTemplate_ supports wildcards just fine.
I did the following to get the wildcard support fully running in the _PathTemplateMatcher_:

1. added corresponding tests
1. changed  _PathTemplate.create_ method to create a Part with **'*'**
1. changed _PathTemplateMatcher.match_ to consider wildcards correctly
1. changed _PathTemplateMatcher.trimBase_ to cut away **'*'**

### DIFF
--- a/core/src/main/java/io/undertow/util/PathTemplate.java
+++ b/core/src/main/java/io/undertow/util/PathTemplate.java
@@ -45,7 +45,7 @@ public class PathTemplate implements Comparable<PathTemplate> {
     private final String templateString;
     private final boolean template;
     private final String base;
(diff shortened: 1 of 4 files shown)


undertow-pull-request:

stuartwdouglas:
Merged, thanks

PR #457, opened 2016-12-12
[UNDERTOW-930] Setting default encoding doesn't affect HttpServletRequest.characterEncoding

Ir2Sora:
https://issues.jboss.org/projects/UNDERTOW/issues/UNDERTOW-930


undertow-pull-request:

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
```diff
@@ -315,9 +316,19 @@ public String getDefaultEncoding() {
...
         return this;
     }
 
+    /**
+     * Determine whether default encoding was explicitly set
+     *
+     * @return true if default encoding was explicitly set
+     */
+    public boolean isDefaultEncodingExplicitlySet() {
+        return defaultEncodingExplicitlySet;
```
Ir2Sora: I did this to keep backward compatibility. Now defaultEncoding in DeploymentInfo is nullable and check for explicit defaultEncoding set implemented through comparison with null

stuartwdouglas:
Squashed and merged, thanks

PR #463, opened 2016-12-19
Avoid a ConcurrentModificationException

jwendell:
By making a defensive copy of the receivers set.

The r.markStreamBroken() method calls back the notifyClosed()
method which removes an element from the set. This cannot be done
within an interaction over the set.


jwendell:
I can cherry-pick this one and open a PR against master as well. OK to do so?

undertow-pull-request:

stuartwdouglas:
Thanks, I have given this the JIRA https://issues.jboss.org/browse/UNDERTOW-938 and applied your fix to all three branches.

PR #472, opened 2017-01-16
Allow custom socket options per listener

cakofony:
I would like to require certificates for some listeners but not others, without shedding the builder api.

### DIFF
--- a/core/src/main/java/io/undertow/Undertow.java
+++ b/core/src/main/java/io/undertow/Undertow.java
@@ -152,7 +152,8 @@ public final class Undertow {
                     AjpOpenListener openListener = new AjpOpenListener(buffers, serverOptions);
                     openListener.setRootHandler(rootHandler);
                     ChannelListener<AcceptingChannel<StreamConnection>> acceptListener = ChannelListeners.openListenerAdapter(openListener);
-                    AcceptingChannel<? extends StreamConnection> server = worker.createStreamConnectionServer(new InetSocketAddress(Inet4Address.getByName(listener.host), listener.port), acceptListener, socketOptions);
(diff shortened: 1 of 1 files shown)


undertow-pull-request:

core/src/main/java/io/undertow/Undertow.java
```diff
@@ -250,14 +253,15 @@ public XnioWorker getWorker() {
         AJP
     }
 
-    private static class ListenerConfig {
+    public static class ListenerConfig {
```
stuartwdouglas: Why did you make this public? It should not be part of the API

stuartwdouglas: I think it would be better to add an Undertow.ListenerBuilder class, and just expose that as the public API, then have addListener(ListenerBuilder). I just don't really like the idea of a user having to deal with 3 levels of nestled classes (new Undertow.ListenerConfig.Builder())

cakofony: It's the argument to `Undertow.Builder.addListener`. Alternatively I could build a similar but slightly differnet model object, but it would be more or less identical. Any suggestions?

cakofony: Sounds good, thanks! I'll knock it out later tonight!

cakofony:
Updated

cakofony:
Thanks for the review @stuartwdouglas! Is there a release schedule you could point me to? I'm looking forward to taking advantage of this change.

stuartwdouglas:
1.4.9 will be released some time in the next two weeks.

Stuart

PR #477, opened 2017-02-01
UNDERTOW-853 SPDY: SpdyClientConnection.sendRequest() shouldn't be synchronized

jwendell:
Revert "SPDY Connection: Make synchronized requests"

This reverts commit 595322a869494ba03610773b2e8084da8d4a8372.

### DIFF
--- a/core/src/main/java/io/undertow/client/spdy/SpdyClientConnection.java
+++ b/core/src/main/java/io/undertow/client/spdy/SpdyClientConnection.java
@@ -85,7 +85,7 @@ public class SpdyClientConnection implements ClientConnection {
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


undertow-pull-request:

jwendell:
ping

ctomc:
this is ok to test

stuartwdouglas:
The issue is that without the sync the requests can be sent out of order which will break the connection. 

This really needs a fix similar to https://github.com/undertow-io/undertow/commit/91417fa220ccf9b58968b2505d0bb34086e2c622, however I am not sure when I will get time to look at it (SPDY is a dying protocol, so it is not high on my priority list)

jwendell:
OK, let's close this one and track https://github.com/undertow-io/undertow/pull/512 instead.

PR #486, opened 2017-02-11
UNDERTOW-987 Providing Java 9 variant of FastConcurrentDirectDeque which doesn't use Unsafe

gunnarmorling:
Hey, this could need some more testing, but before taking it any further I wanted to gauge interest whether you are interested in it at all at this point or not (I took a look into this primarily to gain some more experience with Java 9 and the required changes).

On the change itself: `FastConcurrentDirectDeque` uses `Unsafe` which should be prevented as of Java 9. So I updated it based on the latest `ConcurrentLinkedDequeue` which has moved to var handles already.

The [Multi-Release JAR mechanism](http://openjdk.java.net/jeps/238) is used to provide that new version of `FCDD` to Java 9 users while exposing the one based on `Unsafe` otherwise. The version for Java 8 has been updated as per the `CLD` in the latest JDK 8 (I've re-copied it from the upstream JSR 166 repo and re-applied Jason's extensions).

Tool support for building MR JARs is essentially non-existent at this point. I'm calling javac another time via the maven-antrun-plug-in to compile the Java 9 specific code. It is then added to _META-INF/versions/9_ in the resulting JAR. The Java 9 specific sources are under _src/main/java9_ and you'll get a duplicate class error when having this and _src/main/java_ as source folders in your IDE. But IMO that's preferable over having that code in another module and the circular dependencies implied by that (usually you won't need _src/main/java9_ in your IDE anyways).


undertow-pull-request:
**Windows** [Build 1952](https://ci.wildfly.org/viewLog.html?buildId=45373&buildTypeId=UNDERTOW_MasterWindows) outcome was **FAILURE** using a merge of 7f1acb97ed634c389a9b8357e0c85bf9a3cabff3
Summary: Tests failed: 1, passed: 2744, ignored: 556 Build time: 00:33:36
### Failed tests
```
io.undertow.server.handlers.ReceiverTestCase.testAsyncReceivePartialBytes: java.lang.RuntimeException: test failed with i equal to 6
	at io.undertow.server.handlers.ReceiverTestCase.doTest(ReceiverTestCase.java:233)
	at io.undertow.server.handlers.ReceiverTestCase.testAsyncReceivePartialBytes(ReceiverTestCase.java:200)
Caused by: java.net.SocketTimeoutException: Read timed out
	at io.undertow.server.handlers.ReceiverTestCase.runTest(ReceiverTestCase.java:246)
	at io.undertow.server.handlers.ReceiverTestCase.doTest(ReceiverTestCase.java:231)
	at io.undertow.server.handlers.ReceiverTestCase.testAsyncReceivePartialBytes(ReceiverTestCase.java:200)


```

stuartwdouglas:
Thanks. Do you think it is worth porting this to 1.4.x? I don't really want to do the releases using JDK9 till it is final, so I was thinking it may be worth holding off backporting until it is released.

gunnarmorling:
Good question, what's the reason for the major jump to 2.x? Holding off the Java 9 bits seems reasonable, I also wouldn't fancy building a release using a non-Final Java version. Though for an Alpha release it may be acceptable, and this may be interesting to facilitate Java 9 testing of others.

stuartwdouglas:
2.x will be Servlet 4.0.

PR #498, opened 2017-03-20
UNDERTOW-1023 Respond 400 Bad Request and log at DEBUG level when exc…

msfm:
…eeding max-cookies limit

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/ajp/AjpReadListener.java
+++ b/core/src/main/java/io/undertow/server/protocol/ajp/AjpReadListener.java
@@ -32,6 +32,7 @@ import io.undertow.util.HeaderMap;
 import io.undertow.util.Headers;
 import io.undertow.util.HttpString;
 import io.undertow.util.Methods;
+import io.undertow.util.Cookies;
 import org.xnio.ChannelListener;
 import io.undertow.connector.PooledByteBuffer;
 import io.undertow.util.StatusCodes;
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/undertow/server/protocol/ajp/AjpReadListener.java
```diff
@@ -215,6 +215,17 @@ public void handleEvent(AjpServerResponseConduit channel) {
             //we need to set the write ready handler. This allows the response conduit to wrap it
             responseConduit.setWriteReadyHandler(writeReadyHandler);
 
+            // Check if the number of cookies sent exceeded the maximum
+            try {
+                httpServerExchange.getRequestCookies();
```
stuartwdouglas: This is not a good approach for performance reasons, as this involves creating a cookie map that may not be used (e.g. if the request is just serving a resource and does not care about cookies). 

Any sort of check for this would also have to be dealt with in the HTTP/2 code to keep it consistent.

msfm: Thank you for the review.

> This is not a good approach for performance reasons, as this involves creating a cookie map that may not be used (e.g. if the request is just serving a resource and does not care about cookies).

I'll change to invoke `Cookies.parseRequestCookies()` directly instead of calling it via `httpServerExchange.getRequestCookies()`. This does not create a cookie map. Is this approach ok?

> Any sort of check for this would also have to be dealt with in the HTTP/2 code to keep it consistent.

Got it. I'll also update `Http2ReceiveListener.handleRequests()`, too. 

By the way, 

> ERROR [io.undertow] (default I/O-24) UT005085: Connection io.undertow.server.protocol.http2.Http2ServerConnection for exchange HttpServerExchange{ ...(snip)... } was not closed cleanly, forcibly closing connection

I noticed the above ERROR log happened when max-parameters is exceeded (and max-cookies when I tested). So, I'll also update this PR with moving the try-catch block of `Connectors.setExchangeRequestPath()` after calling 
 `dataChannel.getResponseChannel().setCompletionListener()`.

msfm:
I've just pushed the update of this PR.

undertow-pull-request:
**Linux** [Build 314](https://ci.wildfly.org/viewLog.html?buildId=51967&buildTypeId=UNDERTOW_14xLinux) outcome was **FAILURE** using a merge of b1257b4c99ff793adaaa26d52c17b4f10eb91cfa
Summary: Execution timeout (new); tests passed: 2756, ignored: 478 Build time: 01:00:07

msfm:
I updated this PR again because it was conflicted with the latest change. Are there any additional changes you'd like me to make on this PR? Of course, totally fine with me if you prefer to modify it yourself.

---

As per your review, I made the following changes on this PR:

 - Invoke `Cookies.parseRequestCookies()` directly instead of calling it via `httpServerExchange.getRequestCookies()`
   (I think this approach does not create cookie map. If this approach is still not a good apprpoach, please point it out.)
 - Include the same checking in `Http2ReceiveListener.handleRequests()` for HTTP/2 connector, too.

In addition to the above, I made the following changes:
 
 - Add UndertowOptions.DEFAULT_MAX_COOKIES for the default value
 - Move the try-catch block of `Connectors.setExchangeRequestPath()` to after calling 
 `dataChannel.getResponseChannel().setCompletionListener()` in `Http2ReceiveListener.handleRequests()` because the following ERROR happened when I tested exceeding max-cookies and max-parameters.

    > ERROR [io.undertow] (default I/O-24) UT005085: Connection io.undertow.server.protocol.http2.Http2ServerConnection for exchange HttpServerExchange{ ...(snip)... } was not closed cleanly, forcibly closing connection

msfm:
I'm closing this PR at this moment. I'll resolving a conflict, then I'll send a new PR to the master branch instead.

PR #500, opened 2017-03-24
UndertowHttp2Client and test cases

stevehu:
There are still some debug info shows up but all test cases are executed successfully. Let me know if anything needs to be updated. Thanks.

### DIFF
--- a/core/src/main/java/io/undertow/client/UndertowHttp2Client.java
+++ b/core/src/main/java/io/undertow/client/UndertowHttp2Client.java
@@ -1,27 +1,9 @@
-/*
- * JBoss, Home of Professional Open Source.
- * Copyright 2014 Red Hat, Inc., and individual contributors
- * as indicated by the @author tags.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- *  Unless required by applicable law or agreed to in writing, software
- *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


undertow-pull-request:

core/src/main/java/io/undertow/client/UndertowHttp2Client.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.Map;
+import java.util.ServiceLoader;
+
+/**
+ * Undertow http/2 client class. This class loads {@link ClientProvider} implementations, and uses them to
+ * create connections to a target.
+ *
+ * @author Steve Hu
+ */
+public class UndertowHttp2Client {
```
stuartwdouglas: What is the purpose of this class? The standard HTTP client can handle HTTP/2 connections?

stevehu:
I debugged into the UndertowClient and found that it can only load HTTP/1.1 ClientProviders with the key of "http" or "https". HTTP/2 providers have key as "h2", "h2c" and "h2-prior" and they cannot be loaded based from the following line.

```
ClientProvider provider = clientProviders.get(uri.getScheme());
```

stuartwdouglas:
If you want to force h2 then you need to use URI's such as 'h2://localhost:8443', however if you pass the ENABLE_HTTP2 option into the client then the HTTP/1.1 provider will attempt to upgrade to H2 (see the impl at https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/client/http/HttpClientConnection.java#L139).

stevehu:
Great!!!! Thanks Stuart for the clarification. I missed the upgrade logic when reading the code.

stuartwdouglas:
I am still happy to include the test, I just don't really think we need a separate HTTP/2 client.

stevehu:
Cool. I will added some test cases to the existing one and give you another pull request. Thanks.

PR #510, opened 2017-05-18
Fix potential infinite loop in AbstractFramedChannel#flushSenders

ossnms:
Occasionally I/O threads would get stuck in an infinite loop, throwing the following exception over and over:

`ERROR [default I/O-8] [org.xnio.listener] XNIO001007: A channel event listener threw an exception: java.lang.IllegalStateException: XNIO000017: Buffer was already freed
	at org.xnio.ByteBufferSlicePool$PooledByteBuffer.getResource(ByteBufferSlicePool.java:206)
	at org.xnio.ByteBufferSlicePool$PooledByteBuffer.getResource(ByteBufferSlicePool.java:176)
	at io.undertow.server.protocol.framed.AbstractFramedChannel.flushSenders(AbstractFramedChannel.java:481)
	at io.undertow.server.protocol.framed.AbstractFramedChannel$FrameWriteListener.handleEvent(AbstractFramedChannel.java:774)
	at io.undertow.server.protocol.framed.AbstractFramedChannel$FrameWriteListener.handleEvent(AbstractFramedChannel.java:771)
	at org.xnio.ChannelListeners.invokeChannelListener(ChannelListeners.java:92)
	at org.xnio.conduits.WriteReadyHandler$ChannelListenerHandler.writeReady(WriteReadyHandler.java:65)
	at org.xnio.nio.NioSocketConduit.handleReady(NioSocketConduit.java:93)
	at org.xnio.nio.WorkerThread.run(WorkerThread.java:539)
`

A change was made -- in the line of an earlier commit -- to actually handle such exceptions in AbstractFramedChannel#flushSenders. Hopefully this will then close the channel and prevent an infinite loop.

The underlying root cause could not be determined. The problem can not be reproduced at will.

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
@@ -585,7 +585,7 @@ public abstract class AbstractFramedChannel<C extends AbstractFramedChannel<C, R
                     }
                 }
 
-            } catch (IOException e) {
+            } catch (IOException|RuntimeException e) {
                 safeClose(channel);
                 markWritesBroken(e);
             }


stuartwdouglas:
1.2.x is no longer maintained. Is it possible to reproduce this error on 1.4.x (there has been a lot of changes to this code since 1.2)?

In general that method should not be throwing an exception (except IOException), I would rather deal with the underlying issue if possible.

vveloso:
I agree that the underlying cause should be understood and dealt with. However this is clearly a concurrency problem of difficult reproduction on a complex system such as Wildfly, which we are using.

You committed a similar fix back in 2015 in [Fix potential infinite loop in AbstractFramedChannel](https://github.com/undertow-io/undertow/commit/0c82b05d4d4e9ae143f67d0c37ea186c0e4cf0b9) which applied the exact same workaround and unfortunately apparently addressed only one part of the same problem.

We will be migrating away from Wildfly 8 in a couple of months. Only then we'll see if the problem is still present in later versions.

PR #512, opened 2017-05-30
UNDERTOW-1088 SpdyClientConnection.sendRequest should not be syncronized

jwendell:


jwendell:
I borrowed the same concept applied to http2. In my tests it worked fine.

PR #513, opened 2017-05-30
Explicit surefire plugin version

rhatlapa:
Lets have defined explicit surefire plugin version to prevent usage of different versions on different environments.


ctomc:
This change doesn't really do much, as version is managed in jboss-parent in any case....
but it doesn't hurt

PR #515, opened 2017-05-31
Updated jacoco version to 0.7.9 + merged jacoco report

rhatlapa:
Lets put the jacoco update also to 1.4.x branch


stuartwdouglas:
I had already back ported this locally just forgot to push it

PR #526, opened 2017-06-13
UNDERTOW-1085 Fix NullPointerException

webfolderio:
If DeploymentInfo.securityDisabled is true ServletPathMatches throws NullPointerException.


undertow-pull-request:

stuartwdouglas:
This will not correctly apply wrappers if security is disabled

PR #534, opened 2017-07-18
UNDERTOW-1139 Not-accepted websocket extension will NPE connection

favila:
If a websocket ExtensionHandshake is available but fails negotiation against the client-supplied extensions, the websocket connection will die with a null pointer exception.

The problem is that the Handshake class does not check the result of `ExtensionHandshake.accept(WebSocketExtension ext)` for null (which means the extension should not be used) before adding this extension to the list of selected and configured extensions on the connection. The null blows up later.

This commit checks whether the available extension was accepted by checking for a null return value from the extension's `accept` method. It stops checking for `ext != null` because this is unnecessary.


undertow-pull-request:

PR #539, opened 2017-07-31
[UNDERTOW-1144] URL session tracking - sessionid is appended if there is old sessionid in the URL

iweiss:
Issue: https://issues.jboss.org/browse/UNDERTOW-1144


stuartwdouglas:
This falls over if there are path parameters in addition to jsessionid, or if any of the other additional parameters contains jsessionid as part of the name or the value of the parameter

PR #542, opened 2017-08-09
[UNDERTOW-1149] Move CachedAuthenticatedSessionMechanism before than NegotiationMechanism to avoid kerberos negotiation on each request in IE

yersan:
IE (at least IE8, IE9 and IE10) sends the negotiation token in every request for kerberos authentication. This is only reproducible in undertow-1.3, because in undertow-1.4 CachedAuthenticatedSessionMechanism is being called before than NegotiationMechanism. This patch fixes the order of those mechanisms in undertow-1.3 as they are in undertow-1.4. 

Issue: https://issues.jboss.org/browse/UNDERTOW-1149
EAP 7.0.x Issue: https://issues.jboss.org/browse/JBEAP-12407
Upstream Issue: Upstream not required


undertow-pull-request:

PR #543, opened 2017-08-10
FIX: Only send path (part after hostname) to upstream for HTTP proxy requests

ljwagerfield:
When using `ProxyHandler` to build a forward proxy, the request path forwarded to the upstream is incorrect. It does not correctly strip-out the `scheme://hostname` prefix. Instead, it only strips-out the `scheme:` prefix.

For example, the following request to the Undertow server:
```
GET http://github.com/undertow-io HTTP/1.1
...
```

Incorrectly produces the following request to the upstream:
```
GET //github.com/undertow-io HTTP/1.1
...
```

The fix in the PR corrects this to:
```
GET /undertow-io HTTP/1.1
...
```


undertow-pull-request:

stuartwdouglas:
Merged under https://issues.jboss.org/browse/UNDERTOW-1153

PR #546, opened 2017-08-22
[master branch] UNDERTOW-1016 Fix potential NPE while dealing with unresolved InetSocketAddress

jaikiran:
The commit here fixes the issue reported in https://issues.jboss.org/browse/UNDERTOW-1016.

The commit just reorders the if blocks which check whether an address is unresolved and whether it is a "any" local address, such that it first checks whether the address is unresolved. This prevents the NPE while checking `isAnyLocalAddress()` since the `managementSocketAddress.getAddress()`, as per its API contract can return null for unresolved address.


rhusar:
Thanks for the fix @jaikiran !

PR #547, opened 2017-08-23
HttpClientConnection catches failures more broadly

cakofony:
In some (likely failure) cases I've noticed the undertow client
fails to execute any ClientCallback methods.

I've noticed NPEs in the the logs from SslConduit.doUnwrap, so
it's possible this is caused by UNDERTOW-1156.

### DIFF
--- a/core/src/main/java/io/undertow/client/http/HttpClientConnection.java
+++ b/core/src/main/java/io/undertow/client/http/HttpClientConnection.java
@@ -126,7 +126,6 @@ class HttpClientConnection extends AbstractAttachable implements Closeable, Clie
     private static final int UPGRADE_REQUESTED = 1 << 29;
     private static final int CLOSE_REQ = 1 << 30;
     private static final int CLOSED = 1 << 31;
-    private int count = 0;
 
     private int state;
     private final ChannelListener.SimpleSetter<HttpClientConnection> closeSetter = new ChannelListener.SimpleSetter<>();
@@ -331,7 +330,6 @@ class HttpClientConnection extends AbstractAttachable implements Closeable, Clie
             http2Delegate.sendRequest(request, clientCallback);
             return;
         }
-        count++;
         if (anyAreSet(state, UPGRADE_REQUESTED | UPGRADED | CLOSE_REQ | CLOSED)) {
             clientCallback.failed(UndertowClientMessages.MESSAGES.invalidConnectionState());
             return;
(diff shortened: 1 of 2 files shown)


cakofony:
This change is a bit paranoid, feel free to close out if you think the problem is likely elsewhere.

undertow-pull-request:
**Windows** [Build 2168](https://ci.wildfly.org/viewLog.html?buildId=71875&buildTypeId=UNDERTOW_MasterWindows) outcome was **FAILURE** using a merge of c5790bb621c7c2f2fb7ae7951119caaacdeefed3
Summary: Tests passed: 2881, ignored: 618; exit code 1 Build time: 00:38:00

core/src/main/java/io/undertow/client/http/HttpClientConnection.java
```diff
@@ -110,25 +110,24 @@ public void handleEvent(StreamSourceConduit channel) {
     private static final Logger log = Logger.getLogger(HttpClientConnection.class);
 
     private final Deque<HttpClientExchange> pendingQueue = new ArrayDeque<>();
-    private HttpClientExchange currentRequest;
-    private HttpResponseBuilder pendingResponse;
+    private volatile HttpClientExchange currentRequest;
+    private volatile HttpResponseBuilder pendingResponse;
```
stuartwdouglas: Why the volatile? This client is not designed to be thread safe, and should old be used from the IO thread that is tied to the connection.

cakofony: Was paranoia. Fixed.

undertow-pull-request:
**Windows** [Build 2169](https://ci.wildfly.org/viewLog.html?buildId=71892&buildTypeId=UNDERTOW_MasterWindows) outcome was **FAILURE** using a merge of 640e3ceee5218ec1ad417fbf279ffc5a2dddac6e
Summary: Tests passed: 2881, ignored: 618; exit code 1 Build time: 00:28:05

cakofony:
Once 1.4.19.Final has published I'll run my test suite against that and report whether or not the hang is fixed.
It looks like the version is tagged but unpublished.

cakofony:
Still having requests fail to hit the callback on `1.4.19.Final` using http1.1. The If I force h2c-prior/h2 protocols instead flakes are considerably less common (haven't seen any since upgrading to 1.4.19).

cakofony:
Added a close setter on checkout which triggers a callback as well, but that doesn't seem to do it either.

For context, I've set up timers on my connections which kill the request after ten minutes if nothing has happened.

The entirety of my request data is sent, but the response callback is never called.

stuartwdouglas:
Do you have any information on which exceptions are being thrown from these methods?

cakofony:
@stuartwdouglas Unfortunately I've not been able to get any logging around the cause

cakofony:
@stuartwdouglas After some digging it appears that the hang occurs (at very low frequency) when clients attempt to connect to servers that are in the process of starting up. Likely some unhandled failure when unexpected data is received.

PR #548, opened 2017-08-24
UNDERTOW-1166: make PROXY protocol implementation compliant with spec

spaletta:
This fixes a typo in the PROXY protocol code (ProxyProtocolReadListener.java:38) that eliminated the "UNKNOWN" handling. In addition to that, the state machine for that case was too narrow, requiring at least a space character after the "UNKNOWN".
Apart from fixing these issues, this PR adds test coverage for the "UNKNOWN" handling code and gives some general love to the PROXY protocol code.


undertow-pull-request:

PR #550, opened 2017-08-29
[1.4.x backport] Attempt to recover from Errors and RuntimeExceptions

cakofony:


cakofony:
Looks like this is already backported, thanks!

PR #553, opened 2017-08-29
Add an example for how to programatically setup SSL with a reasonable…

areese:
… cipher set

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -66,7 +66,11 @@
             <groupId>org.jboss.logmanager</groupId>
             <artifactId>jboss-logmanager</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava </artifactId>
+            <version>22.0</version>
+         </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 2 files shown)


undertow-pull-request:

examples/pom.xml
```diff
@@ -66,7 +66,11 @@
             <groupId>org.jboss.logmanager</groupId>
             <artifactId>jboss-logmanager</artifactId>
         </dependency>
-
+        <dependency>
```
stuartwdouglas: I don't really want to bring in a dependency on guava for just one utility method

examples/src/main/java/io/undertow/examples/ssl/HelloWorldSSL.java
```diff
@@ -0,0 +1,123 @@
...
+@UndertowExample("Hello World SSL")
+public class HelloWorldSSL {
+
+    public static void main(final String[] args) throws Exception {
+        final SSLContext sslContext = SSLContext.getInstance("TLSv1.2");
+        sslContext.init(getKeyManagers(System.getProperty("keyStorePath")),
+                        getTrustManagers(System.getProperty("trustStorePath")), null);
+
+        final List<String> cipherSuites = new ArrayList<>();
+        cipherSuites.add("TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256");
```
stuartwdouglas: Instead of adding a new example I think it may make more sense to just change the HTTP/2 example to use this configuration (as HTTP/2 won't work if a weak cipher is selected)

areese:
Thanks, I'll look at redoing this with http/2.

PR #558, opened 2017-09-11
[Proposal] Threadlocal cache on IO threads, use global queue by default

cakofony:
In servlet applications pooled buffer instances tend to "migrate"
out from IO threads to worker threads, that is to say they are
allocated on an IO thread, but returned from a worker. Using the
default configuration from Undertow.Builder the global pool is
unused, forcing much higher re-allocation than should be necessary.

This commit makes two changes:
1. The ThreadLocal cache is only used on IO threads. This allows
the value to be safely modified to something much higher for faster
IO thread operations without the risk of significant memory bloat
as buffers "migrate" away.
2. The default buffer pool configuration from Undertow.Builder
has been updated to use a ThreadLocal cache size of 12 (the value
used by the default ssl buffer pool) up from 4, and a global
shared pool size matching the worker thread count.

### DIFF
--- a/core/src/main/java/io/undertow/Undertow.java
+++ b/core/src/main/java/io/undertow/Undertow.java
@@ -146,7 +146,7 @@ public final class Undertow {
 
             ByteBufferPool buffers = this.byteBufferPool;
             if (buffers == null) {
-                buffers = new DefaultByteBufferPool(directBuffers, bufferSize, -1, 4);
+                buffers = new DefaultByteBufferPool(directBuffers, bufferSize, workerThreads, 12);
             }
 
             listenerInfo = new ArrayList<>();
--- a/core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
+++ b/core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
@@ -21,6 +21,7 @@ package io.undertow.server;
 import io.undertow.UndertowMessages;
 import io.undertow.connector.ByteBufferPool;
 import io.undertow.connector.PooledByteBuffer;
+import org.xnio.XnioIoThread;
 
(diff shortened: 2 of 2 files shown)


cakofony:
Feel free to close this out if it seems like a bad idea. It would mean more contention for buffers on operations running from worker threads.

stuartwdouglas:
I am open to making this configurable, but I don't think it should be the default.

I am also curious as to the 'migration' that you are seeing. Is it just for POST requests?
In general when using Servlet the allocation pattern should be:
- IO Thread: allocate one buffer to read the request in the IO thread, in a GET request this should be returned immediatly, in a POST request it will migrate to a worker
- Worker Thread: allocate one buffer to read the request (for POST requests only). Allocate two buffers to write the request (one for the header, one for the content)

stuartwdouglas:
So thinking about it 2 would be a more appropriate thread local cache size than 4, however the io.undertow.server.DefaultByteBufferPool.ThreadLocalData#allocationDepth field is supposed to limit the TL cache size on threads that only release buffers and do not allocate. 

Basically this tracks allocations from the current thread, and will only return a buffer to the thread local queue if there have been a similar number of allocations from that thread. This is supposed to prevent the buffer migration that you are talking about above.

cakofony:
> Is it just for POST requests?

I believe so, I'm using a post-heavy remoting framework. Using WriteListeners off the IO thread causes the migration as well.

> io.undertow.server.DefaultByteBufferPool.ThreadLocalData#allocationDepth field is supposed to limit the TL cache size on threads that only release buffers and do not allocate

Ah, wonderful, it sounds like this is unnecessary since the buffer pool is configurable. I'm curious if you have thoughts on enabling the shared queue pool by default.

PR #559, opened 2017-09-23
UNDERTOW-1147 Fix

sermojohn:
Added a unit test that exposes this issue. Implemented the fix in the MultiPartParseDefinition, so the unit test succeeds if executed on the patched version.

The unit test could be improved by enforcing the http client to include the "filename*" value in the content-disposition header, but I could not make it. However, this header format is valid based on 
RFC 2231.


undertow-pull-request:

PR #567, opened 2017-10-16
UNDERTOW-1197 Fix proposal

sermojohn:


undertow-pull-request:

servlet/src/main/java/io/undertow/servlet/handlers/FilterHandler.java
```diff
@@ -106,8 +106,8 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
 
 
             final ServletRequestContext servletRequestContext = exchange.getAttachment(ServletRequestContext.ATTACHMENT_KEY);
-            final ServletRequest oldReq = servletRequestContext.getServletRequest();
-            final ServletResponse oldResp = servletRequestContext.getServletResponse();
+            final ServletRequest oldReq = servletRequestContext.getOriginalRequest();
+            final ServletResponse oldResp = servletRequestContext.getOriginalResponse();
```
stuartwdouglas: I really don't think this is correct. If you have multiple filters that wrap the request it will be reset to the original one after the last filter in the chain returns.

sermojohn: I get your point, this is not correct.

However, the reset of request and response reference on the ServletRequestContext does not work well with some filters (like org.springframework.web.filter.ShallowEtagHeaderFilter) that wrap the response in the initial filter execution and require a response of the wrapped type when the asynchronous request completes.

I believe that the ServletRequestContext request and response reset in the finally block (https://github.com/sermojohn/undertow/blob/d37402476127c50ace20a340f64718bc8ba17d90/servlet/src/main/java/io/undertow/servlet/handlers/FilterHandler.java#L144) could be totally removed. This would result in leaking the response wrappers, but this should not affect the filters themselves because they only use the request and response arguments they receive.

What do you think?

stuartwdouglas: I think that the correct way to handle this is to store the current request/response when the AsyncContext is created.

undertow-pull-request:
**Linux** [Build 512](https://ci.wildfly.org/viewLog.html?buildId=79315&buildTypeId=UNDERTOW_14xLinux) outcome was **FAILURE** using a merge of fc686d9ca087ffe8deb3c06395eb2bf56db8fece
Summary: Execution timeout (new); tests passed: 2971, ignored: 559 Build time: 01:00:08

fl4via:
The fix suggested by @stuartwdouglas was provided here #880 
I'm closing this one now.

PR #568, opened 2017-10-20
Add option to enable default websocket deployment info.

afspear:
This would be helpful if using undertow to serve a vaadin application with atmosphere.

So in the server definition you could say:
```
           DeploymentInfo servletBuilder = deployment()
              .setClassLoader(Main.class.getClassLoader())
              .setContextPath(MYAPP)
              .setDeploymentName("vaadin.war")
              .setResourceManager(new FileResourceManager(new File("src/main/webapp"), 1024))
              .enableDefaultWebsocket(true)
              .addServlets(
                servlet("MyUIServlet", MyUI.MyUIServlet.class)
                  .setAsyncSupported(true)
                  .addMapping("/*"));
```

instead of this adding boilerplate like so:

```
            DeploymentInfo servletBuilder = deployment()
              .setClassLoader(Main.class.getClassLoader())
              .setContextPath(MYAPP)
              .setDeploymentName("vaadin.war")
              .setResourceManager(new FileResourceManager(new File("src/main/webapp"), 1024))
              .addServlets(
                servlet("MyUIServlet", MyUI.MyUIServlet.class)
                  .setAsyncSupported(true)
                  .addMapping("/*"));

            WebSocketDeploymentInfo webSocketDeploymentInfo = new WebSocketDeploymentInfo();
            servletBuilder.addServletContextAttribute(WebSocketDeploymentInfo.ATTRIBUTE_NAME, webSocketDeploymentInfo);
```

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
+++ b/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
@@ -126,6 +126,7 @@ public class DeploymentInfo implements Cloneable {
     private final Map<String, AuthenticationMechanismFactory> authenticationMechanisms = new HashMap<>();
     private final List<LifecycleInterceptor> lifecycleInterceptors = new ArrayList<>();
     private final List<SessionListener> sessionListeners = new ArrayList<>();
+    private final boolean defaultWebSocketEnabled = false;
 
     /**
      * additional servlet extensions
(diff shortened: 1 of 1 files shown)


undertow-pull-request:

stuartwdouglas:
This would need to use some kind of service loader based approach to make this work as websockets depends on Servlet and not the other way around. Given that the boilderplate is only a single line (no need to assign it to a variable first ) I don't think this this woth it.

PR #580, opened 2017-10-31
Avoid mandatory dependency to java.sql module

gunnarmorling:
Hey, I was experimenting with [using Undertow as a module on Java 9](https://github.com/moditect/moditect#example) (explicit module, not automatic) and creating a modular runtime image with it.

In that course I noticed that there currently is a mandatory dependence to the _java.sql_ module due to a specific log handler and exposure of `SQLException` in a few signatures of `UndertowLogger`. This proposed change allows to make that dependence optional (`requires static java.sql` in terms of the module descriptor), which reduces size of a runtime image with an Undertow Hello World from 29 MB to 25 MB. The price is a slightly less explicit signature of these methods. An alternative would be to create a separate logger interface just for the methods with a `SQLException` parameter.

I thought I'd first propose this change here quickly and see whether you'd be willing to apply it. Happy to file a JIRA issue if you think it's good. Thanks!

### DIFF
--- a/core/src/main/java/io/undertow/UndertowLogger.java
+++ b/core/src/main/java/io/undertow/UndertowLogger.java
@@ -39,7 +39,6 @@ import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.net.URI;
 import java.nio.file.Path;
-import java.sql.SQLException;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


stuartwdouglas:
This looks good to me, can you file a JIRA?

gunnarmorling:
Cool; filed [UNDERTOW-1223](https://issues.jboss.org/browse/UNDERTOW-1223) and altered the commit message to refer to that issue.

undertow-pull-request:
**Linux** [Build 502](https://ci.wildfly.org/viewLog.html?buildId=78944&buildTypeId=UNDERTOW_14xLinux) outcome was **FAILURE** using a merge of d28738c6f52d4113a12b1701179ec39ee70c1e48
Summary: Execution timeout (new); tests passed: 2970, ignored: 559 Build time: 01:00:14

PR #583, opened 2017-11-04
Proposal to add Handler for Curl format request including body data

iamdvr:
Two enhancements on top of existing RequestDumpHandler

#1) able to handle request body, Proposed one would handle it.
#2) Current one prints in regular format, which cannot be copied and replicate same request in bash/terminal. We try to print in curl acceptable format, so it should be quick copy/paste.

Here is sample Log line:

CURL Request Format --> curl 'http://127.0.0.1:8080/testcontextpath/test' --cookie '_ga=GA1.1.16270314.1505902496' -H 'Accept: application/json' -H 'Accept-Language: en-US,en;q=0.8' -H 'Accept-Encoding: gzip, deflate, br' -H 'Origin: http://127.0.0.1:8080' -H 'User-Agent: Mozilla/5.0' -H 'Connection: keep-alive' -H 'Content-Length: 17' -H 'Content-Type: application/json;charset=UTF-8' -H 'Referer: http://127.0.0.1:8080/testcontextpath/' -H 'Host: 127.0.0.1:8080' --data-binary '{"someDatId":"1"}'

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/CurlRequestDumpingConduit.java
+++ b/core/src/main/java/io/undertow/server/handlers/CurlRequestDumpingConduit.java
@@ -1,0 +1,80 @@
+package io.undertow.server.handlers;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.xnio.IoUtils;
+import org.xnio.channels.StreamSinkChannel;
+import org.xnio.conduits.AbstractStreamSourceConduit;
+import org.xnio.conduits.ConduitReadableByteChannel;
+import org.xnio.conduits.StreamSourceConduit;
+
+public class CurlRequestDumpingConduit extends AbstractStreamSourceConduit<StreamSourceConduit>
+{
+
+    private final List<byte[]> data = new CopyOnWriteArrayList<>();
+
+    /**
+     * Construct a new instance.
+     *
+     * @param next the delegate conduit to set
+     */
+    public CurlRequestDumpingConduit(StreamSourceConduit next)
+    {
+        super(next);
+    }
+
+    public long transferTo(final long position, final long count, final FileChannel target) throws IOException
+    {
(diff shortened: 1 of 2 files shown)


undertow-pull-request:

iamdvr:
Could someone look at it

core/src/main/java/io/undertow/server/handlers/CurlRequestDumpingHandler.java
```diff
@@ -0,0 +1,114 @@
...
+                    return factory.create();
+                }
+            });
+
+            exchange.addExchangeCompleteListener(new ExchangeCompletionListener()
+            {
+                @Override
+                public void exchangeEvent(final HttpServerExchange exchange, final NextListener nextListener)
+                {
+                    sb.append(" --data-binary '" + new String(dumpConduit.fullRequestDump()) + "'");
```
stuartwdouglas: This seems really problematic. In particular:
- What happens if the data starts with an @ character
- What happens if the data contains a ' character
- What happens if the data contains other non-ascii characters that will not be handled correctly?

I think you need to base64 encode this data and pipe it to curl. Something like:

echo '$base64encodeddata' | base64 | curl --data-binary @- ...

Rather than getting the binary data as a string you need to base64 encode the binary data. Of course if there is no data all this should be omitted and it should just call curl directly.

core/src/main/java/io/undertow/server/handlers/CurlRequestDumpingConduit.java
```diff
@@ -0,0 +1,80 @@
...
+import org.xnio.IoUtils;
+import org.xnio.channels.StreamSinkChannel;
+import org.xnio.conduits.AbstractStreamSourceConduit;
+import org.xnio.conduits.ConduitReadableByteChannel;
+import org.xnio.conduits.StreamSourceConduit;
+
+public class CurlRequestDumpingConduit extends AbstractStreamSourceConduit<StreamSourceConduit>
+{
+
+    private final List<byte[]> data = new CopyOnWriteArrayList<>();
```
billoneil: Since this is created per request I don't think it needs to be thread safe. A normal list should be fine.

fl4via:
Closing as the repository has been deleted.

PR #584, opened 2017-11-05
UNDERTOW-1227 Proposed fix

sermojohn:
This fix was tested with a simple spring-boot application that defines a POST controller accepting multi-part requests. No additional unit tests were written.
I implemented a generic type for holding the getParts() and parseFormData() methods output, holding the raised exception when appeared, or the actual result.

### DIFF
--- a/core/src/main/java/io/undertow/UndertowMessages.java
+++ b/core/src/main/java/io/undertow/UndertowMessages.java
@@ -95,7 +95,7 @@ public interface UndertowMessages {
 //    IOException requestEntityWasTooLarge(SocketAddress address, long size);
 
     @Message(id = 20, value = "Connection terminated as request was larger than %s")
-    IOException requestEntityWasTooLarge(long size);
+    IllegalStateException requestEntityWasTooLarge(long size);
 
     @Message(id = 21, value = "Session already invalidated")
     IllegalStateException sessionAlreadyInvalidated();
--- a/core/src/main/java/io/undertow/util/Result.java
+++ b/core/src/main/java/io/undertow/util/Result.java
@@ -1,0 +1,39 @@
+package io.undertow.util;
+
+public class Result<T, X extends RuntimeException> {
+
+    private T result;
+    private X exception;
+
+    public Result(T result, X exception) {
+        this.result = result;
+        this.exception = exception;
+    }
+
+
+    public boolean hasResult() {
+        return result != null;
+    }
+
+    public X getException() {
+        return exception;
+    }
+
+    public T getOrThrow() throws X {
+        if (result == null && exception == null) {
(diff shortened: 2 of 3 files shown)


undertow-pull-request:

stuartwdouglas:
This is ok to test

core/src/main/java/io/undertow/UndertowMessages.java
```diff
@@ -95,7 +95,7 @@
 //    IOException requestEntityWasTooLarge(SocketAddress address, long size);
 
     @Message(id = 20, value = "Connection terminated as request was larger than %s")
-    IOException requestEntityWasTooLarge(long size);
+    IllegalStateException requestEntityWasTooLarge(long size);
```
stuartwdouglas: This is not really correct, this exception can be thrown by any IO operation (not just form parsing), and it is not expected that read() operations will thrown IllegalStateException.

stuartwdouglas:
Thanks for the PR, however I have fixed this a different way (and also fixed the other cases that should throw IllegalStateException)

patrickbray:
Hi @stuartwdouglas & @sermojohn I have arrived at this PR from this Stack Overflow [issue]( https://stackoverflow.com/questions/46968098/undertow-throws-runtimeexception-when-uploading-multipart-file-exceeding-the-set). I see that the underlying issue UNDERTOW-1227 has been resolved in `1.4.22.Final` with a slightly different change to the PR then what was presented here. 

After upgrading to Spring Boot version `1.5.10.RELEASE` and associated embedded undertow version I am still unable to catch the `java.lang.IllegalStateException: io.undertow.server.RequestTooBigException: UT000020: Connection terminated as request was larger than 15728640` inside of a `@ControllerAdvice` is this to be expected?

vinujan59:
> Hi @stuartwdouglas & @sermojohn I have arrived at this PR from this Stack Overflow [issue](https://stackoverflow.com/questions/46968098/undertow-throws-runtimeexception-when-uploading-multipart-file-exceeding-the-set). I see that the underlying issue UNDERTOW-1227 has been resolved in `1.4.22.Final` with a slightly different change to the PR then what was presented here.
> 
> After upgrading to Spring Boot version `1.5.10.RELEASE` and associated embedded undertow version I am still unable to catch the `java.lang.IllegalStateException: io.undertow.server.RequestTooBigException: UT000020: Connection terminated as request was larger than 15728640` inside of a `@ControllerAdvice` is this to be expected?
> 
> Basically what I am keen to do is to return a 400 Bad Request to the user when they attempt to upload a file which exceeds the maximum file size rather then simply closing the underlying connection which seem to be the case currently.

Facing the same issue with spring-boot

PR #588, opened 2017-12-05
UNDERTOW-1237 Added multiline header parsing (master)

cystek:

### DIFF
--- a/core/src/main/java/io/undertow/util/MultipartParser.java
+++ b/core/src/main/java/io/undertow/util/MultipartParser.java
@@ -32,6 +32,12 @@ import io.undertow.connector.PooledByteBuffer;
 public class MultipartParser {
 
     /**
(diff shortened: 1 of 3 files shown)


cystek:
I'm not sure how to create a pull request for specific branches, but this should work with at least 1.4.x as is and 1.3.x with StandardCharsets.UTF_8 removed from the test case.

stuartwdouglas:
Merged, thanks. It is not necessary to open PR's for the older branches, I backport as required.

In most cases this is just a backport to 1.4.x, as 1.3.x is only maintained for EAP7.0 fixes, and older branches are no longer maintained.

cystek:
Ok, thanks. I'll remember that in the future. Will this be likely to be in the 1.4.22 release, then? I've seen header wrapping in a few client libraries already, including Apache HttpComponents.

PR #593, opened 2017-12-06
UNDERTOW-1237 Added multiline header parsing (1.1.x)

cystek:


billoneil:
I don't think you need a PR for every branch. If it's merged they will handle updating existing branches.

cystek:
Oh, I guess the damage is already done but thanks for the info. I wasn't sure if master was only applicable to the 2.0.x version.

stuartwdouglas:
This branch is no longer maintained

PR #594, opened 2017-12-06
UNDERTOW-1237 Added multiline header parsing (1.0.x)

cystek:


stuartwdouglas:
This branch is no longer maintained

PR #597, opened 2017-12-20
RFC per cookie request limit handler

ivagulin:
IMHO RequestLimitingHandler undertow have now is pretty much useless. Same effect is easily achievable by limiting xnio thread pool. Maybe it make sense to add something more usefull: like per cookie limiter? 

I know I have to drop guava and maybe add builder/wrapper (btw wasn't able to figure out who use it). If comminuty is interested I can polish pull-request.

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/CookieRequestLimitingHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/CookieRequestLimitingHandler.java
@@ -1,0 +1,71 @@
+package com.parallels.pa.reqlimit;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.server.handlers.Cookie;
+import io.undertow.server.handlers.RequestLimit;
+
+public class CookieRequestLimitingHandler implements HttpHandler {
+	private int maximumConcurrentRequests = 5;
+	private int queueSize = 10;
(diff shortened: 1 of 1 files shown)


billoneil:
Anyone who is trying to DDoS you or accidentally hit your web server too frequently will probably be doing so programmatically without cookies enabled.

ivagulin:
IMHO it is pretty easy to protect login page. Make every authenticated requests perform well is another story. It is not replacement for full-blown DDoS protection.

For our app scenario is simple: internet explorer give burst of requests if form posted by enter and you press and hold enter. User was on not so fast page when he drop coffee at his keyboard.

stuartwdouglas:
The main difference between RequestLimitingHandler and limiting the pool is that in an environment such as Wildfly the pool can be shared between different services. In addition as Undertow is non blocking not all requests will use a worker thread.

ivagulin:
>Undertow is non blocking not all requests will use a worker thread
I understand this but in the end result is pretty much same: server stop accept requests. I don't think in our "microservices" days lot use several service per app server. My opinion is RequestLimit good piece of code and should have wider use.
Do I have a chance with this per cookie limiter? Should I try to polish it?

stuartwdouglas:
Unless there is widespread interest in this I don't really know if this should be included in the default Undertow jar. The core jar is already quite large and I don't really want to include new handlers unless there is a definite interest.

billoneil:
Although I do think this is very useful it's probably too app specific to be included in Undertow core. You may want it to trigger off of a cookie, I might want it off of a header, someone else from a query param like an API token. Once its app specific you also need to consider how many services are running? Do I want to divide my rate limit by N services or have a shared cache / data store they all use. What if I only want to rate limit anonymous users and not authenticated users. Or maybe different users get different rate limits. In order to cover all those cases you end up with just a `PredicateHandler` and each app will need to customize it to their own tech stack and requirements. Just my two cents.

ivagulin:
Will live in my wildfly module then :(

PR #605, opened 2018-02-05
UNDERTOW-1274 Cross context session creation should not set a cookie,…

jstourac:
… but rely on the original contexts cookie


jstourac:
Update of 'Fix versions' field should be done for  Jira UNDERTOW-1274 if this is merged.

PR #622, opened 2018-03-21
AbstractFramedChannel.outstandingBuffersUpdater is static final

cakofony:
was previously volatile non-static.


cakofony:
Caught while looking into https://issues.jboss.org/browse/UNDERTOW-1312
I don't think it's related to the bug since all invocations are isolated to `this`.

stuartwdouglas:
Thanks, I have committed this under https://issues.jboss.org/browse/UNDERTOW-1313

PR #626, opened 2018-03-24
getUnsafe uses AccessController.doPrivileged

cakofony:
Previously it invoked PrivilegedAction.run directly.

### DIFF
--- a/core/src/main/java/io/undertow/util/FastConcurrentDirectDeque.java
+++ b/core/src/main/java/io/undertow/util/FastConcurrentDirectDeque.java
@@ -1647,11 +1647,11 @@ public class FastConcurrentDirectDeque<E>
 
     private static Unsafe getUnsafe() {
         if (System.getSecurityManager() != null) {
-            return new PrivilegedAction<Unsafe>() {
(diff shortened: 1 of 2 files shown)


stuartwdouglas:
Merged under https://issues.jboss.org/browse/UNDERTOW-1318, thanks

core/src/main/java/io/undertow/util/FastConcurrentDirectDeque.java
```diff
@@ -1647,11 +1647,11 @@ private boolean casTail(Node<E> cmp, Node<E> val) {
 
     private static Unsafe getUnsafe() {
         if (System.getSecurityManager() != null) {
-            return new PrivilegedAction<Unsafe>() {
+            AccessController.doPrivileged(new PrivilegedAction<Unsafe>() {
```
cakofony: Missed a `return` here

cakofony: fix: https://github.com/undertow-io/undertow/pull/627

PR #630, opened 2018-03-28
UNDERTOW-1321: Prevent AbstractFramedChannel deadlock on shutdown

cakofony:
Rather than executing remaining tasks on the current non-io thread,
they are invoked from a fallback single threaded executor.

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/ShutdownFallbackExecutor.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/ShutdownFallbackExecutor.java
@@ -18,9 +18,12 @@
 package io.undertow.server.protocol.framed;
 
 import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * {@link ShutdownFallbackExecutor} wrapper around a single threaded executor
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/server/protocol/framed/ShutdownFallbackExecutor.java
```diff
@@ -0,0 +1,47 @@
...
+ * {@link ShutdownFallbackExecutor} wrapper around a single threaded executor
+ * which will execute pending tasks when the worker has been shut down.
+ */
+final class ShutdownFallbackExecutor {
+    private static volatile Executor EXECUTOR = null;
+    private ShutdownFallbackExecutor() {
+        // Static Utility
+    }
+
+    static void execute(Runnable runnable) {
```
stuartwdouglas: I think this idea is safer, however this executor needs to shut down when it has run out of tasks. Setting the core pool size to zero does not really accomplish this, as the JDK executor does not work the way you would expect it to (it won't grow past core threads until the queue is full, so in this case I don't think this executor will actually function, I don't know if it has special case coding for a core pool size of zero however).

cakofony: This works the same as `Executors.newCachedThreadPool()` with a max size of one rather than `Integer.MAX_VALUE`, and a much shorter keepalive time (100ms down from 1 minute).

Shutting down the executor would require additional synchronization that I'd prefer to avoid if possible.

threads will all be reaped 100ms (though we might as well drop that down to 10ms here) after the last task they execute. Lazily creating the executor is a (likely unnecessary) optimization, we could statically define it.

cakofony:
@stuartwdouglas I've tested this change in my codebase for the past several days, tests no longer deadlock on shutdown (was previously reproducing several times daily).

PR #631, opened 2018-03-29
UNDERTOW-1322 use exchange's sourceAddress instead of getConnection().getPeerAddress() inside ProxyHandler so that it can be used together with ProxyPeerAddressHandler

pfyod:
Fix proposal for https://issues.jboss.org/browse/UNDERTOW-1322


undertow-pull-request:

PR #638, opened 2018-04-21
UNDERTOW-1338 File descriptor leak cause JarURLConnection.getLastModi…

n-agetsu:
Please review for https://issues.jboss.org/browse/UNDERTOW-1338.

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/resource/URLResource.java
+++ b/core/src/main/java/io/undertow/server/handlers/resource/URLResource.java
@@ -21,6 +21,7 @@ package io.undertow.server.handlers.resource;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.JarURLConnection;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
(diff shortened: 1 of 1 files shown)


undertow-pull-request:

core/src/main/java/io/undertow/server/handlers/resource/URLResource.java
```diff
@@ -88,7 +89,13 @@ private void openConnection() {
                     contentLength = null;
                     return;
                 }
-                lastModified =  new Date(connection.getLastModified());
+                if (url.getProtocol().equals("jar")) {
+                    connection.setUseCaches(false);
+                    URL jar = ((JarURLConnection) connection).getJarFileURL();
+                    lastModified = new Date(new File(jar.getFile()).lastModified());
```
stuartwdouglas: Jar files can contain last modified information, so the last modified time of the jar itself may not actually be the last modified time.

stuartwdouglas: Looks like JarURLConnection does not take this into account anyway

PR #639, opened 2018-04-27
UNDERTOW-1342 Improve HttpServletRequestImpl#getLocalAddr() to determ…

msfm:
…ine the IP address from the host name string when destinationAddress is unresolved socket


stuartwdouglas:
Rejecting as per JIRA comment.

PR #642, opened 2018-05-14
UNDERTOW-1346 Add clear methods for collection fields of DeploymentInfo

ppalaga:
https://issues.jboss.org/browse/UNDERTOW-1346


ppalaga:
@stuartwdouglas could you please review?

undertow-pull-request:

stuartwdouglas:
I wonder if it would just be cleaner to remove the unmodifiable collection wrappers so you can just modify them directly.

ppalaga:
We can live with both. If you prefer removing the unmodifiable wrappers I can modify the PR.

stuartwdouglas:
Yea, I think that would be best. There is not really any practical reason they are unmodifiable, just undue caution on my part.

ppalaga:
613fead removes the unmodifiable wrappers.

PR #656, opened 2018-07-12
Use next handler if exists.

eladchen:
When this handler serves a request verb which is not a GET, POST or HEAD
It ends the exchange.

This PR offers a different approach, before ending the exchange, use the next handler in case it is set.
I guess I should be adding a unit test for this — But before I do that, I'd like to know WDYT?

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/resource/ResourceHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/resource/ResourceHandler.java
@@ -159,7 +159,13 @@ public class ResourceHandler implements HttpHandler {
             } else {
                 exchange.setStatusCode(StatusCodes.NOT_IMPLEMENTED);
             }
-            exchange.endExchange();
+
+            if ( this.next != null ) {
+                next.handleRequest(exchange)
+            }
+            else {
+                exchange.endExchange(); 
+            }
         }
     }
 


undertow-pull-request:

core/src/main/java/io/undertow/server/handlers/resource/ResourceHandler.java
```diff
@@ -159,7 +159,13 @@ public void handleRequest(final HttpServerExchange exchange) throws Exception {
             } else {
                 exchange.setStatusCode(StatusCodes.NOT_IMPLEMENTED);
             }
-            exchange.endExchange();
+
+            if ( this.next != null ) {
```
billoneil: This would only get hit if the method is none of the following.

```java
        KNOWN_METHODS.add(Methods.OPTIONS);
        KNOWN_METHODS.add(Methods.GET);
        KNOWN_METHODS.add(Methods.HEAD);
        KNOWN_METHODS.add(Methods.POST);
        KNOWN_METHODS.add(Methods.PUT);
        KNOWN_METHODS.add(Methods.DELETE);
        KNOWN_METHODS.add(Methods.TRACE);
        KNOWN_METHODS.add(Methods.CONNECT);
```

They all get handled above this and will return the method not allowed.

Your code would only trigger if it is a non standard HTTP method. It appears it will also still set the status code to `NOT_IMPLEMENTED` which your handler may or may not be expecting. It will also break the normal functionality of this handler which someone might rely on. It might be better to make this an opt in functionality rather than changing the default behavior.

billoneil: Do you want your logic to replace the inside of https://github.com/undertow-io/undertow/pull/656/files#diff-8c222f88bc42bde6bcc77cb5a15dbb08R155 instead? Then you could add a new handler that defaults to the old functionality and allow someone to override it if they wish. `resourceHandler.setUnsupportedMethodHandler()`

eladchen: I agree. 

So this is the action list:

1. Add a new field of type HttpHandler, called unsupportedMethodHandler; Move the else clause (Line#155) inside of it.
2. A setter ( which verifies the argument isn't null )

eladchen:
Hi, 

The purposed changes have been added, please take a look at let me know what you think.

+ Where should the tests be added?

core/src/main/java/io/undertow/server/handlers/resource/ResourceHandler.java
```diff
@@ -458,7 +455,19 @@ public ResourceHandler setContentEncodedResourceManager(ContentEncodedResourceMa
         return this;
     }
 
-    public boolean isCanonicalizePaths() {
+	public HttpHandler getUnsupportedMethodHandler() {
+		return unsupportedMethodHandler;
+	}
+
+	public void setUnsupportedMethodHandler(HttpHandler unsupportedMethodHandler) {
```
stuartwdouglas: I know that this does not hold true for some of the older handlers, but in general we prefer handlers that are not routing related to be immutable.

core/src/main/java/io/undertow/server/handlers/resource/UnsupportedMethodHandler.java
```diff
@@ -0,0 +1,70 @@
+package io.undertow.server.handlers.resource;
```
stuartwdouglas: Needs the license header

core/src/main/java/io/undertow/server/handlers/resource/UnsupportedMethodHandler.java
```diff
@@ -0,0 +1,70 @@
...
+		exchange.getResponseHeaders().add( Headers.ALLOW, String.join( ", ", Methods.GET_STRING, Methods.HEAD_STRING, Methods.POST_STRING ) );
+	}
+
+	protected void notImplemented( HttpServerExchange exchange ) {
+		exchange.setStatusCode( StatusCodes.NOT_IMPLEMENTED );
+	}
+
+	@Override
+	public void handleRequest( HttpServerExchange exchange ) throws Exception {
+		if ( preferNext && next != null ) {
```
stuartwdouglas: I don't see the point of this. If you want to just delegate to the next handler just do it directly, and don't create a UnsupportedMethodHandler.

core/src/main/java/io/undertow/server/handlers/resource/UnsupportedMethodHandler.java
```diff
@@ -0,0 +1,70 @@
...
+	// Maintain backward compatibility for ResourceHandler (this is package-private on purpose)
+	UnsupportedMethodHandler( final Set<HttpString> notAllowedMethods ) {
+		this.notAllowedMethods = notAllowedMethods;
+		this.preferNext = false;
+		this.next = null;
+	}
+
+	protected void notAllowed( HttpServerExchange exchange ) {
+		exchange.setStatusCode( StatusCodes.METHOD_NOT_ALLOWED );
+		exchange.getResponseHeaders().add( Headers.ALLOW, String.join( ", ", Methods.GET_STRING, Methods.HEAD_STRING, Methods.POST_STRING ) );
```
stuartwdouglas: The allowed methods should not really be hard coded.

core/src/main/java/io/undertow/server/handlers/resource/UnsupportedMethodHandler.java
billoneil: I originally suggested having two handlers similar to how the RoutingHandler has both a `fallbackHandler` and an `invalidMethodHandler`.  The `next` handler seems to be used when the resource is not found such as a 404 handler and this seems like a slightly different use case.

I think this handler could be a private inner class of `ResourceHandler` and never be exposed publicly.

core/src/main/java/io/undertow/server/handlers/resource/ResourceHandler.java
eladchen: What do you mean? You're referring to the "setter" ? Do you suggest this should be done via the constructor ?

core/src/main/java/io/undertow/server/handlers/resource/UnsupportedMethodHandler.java
eladchen: @stuartwdouglas Delegate where? Inside the existing ResourceHandler?

@billoneil 
Why should it not be exposed? What will you gain by keeping it private?
If we do keep this public, someone could reuse the logic, and maybe override the "notAllowed"
& "notImplemented" methods.

Either way, let me know what changes should be made.

core/src/main/java/io/undertow/server/handlers/resource/UnsupportedMethodHandler.java
eladchen: Good catch. I'll make sure these are taken from the "notAllowedMethods" member.

fl4via:
Hello @eladchen!

I am going through the old PRs in Undertow and I found your commit. The goal of the PR is not clear. When would this be used?

fl4via:
Because we have not heard back, I'm closing this one and marking it as invalid. Feel free to reopen it at any time @eladchen and then we will reevaluate it. Thanks!

PR #696, opened 2018-12-04
Regression test for Undertow 1444

mchoma:
Caching resource manager does not handle range requests if file is too large to fit in cache


undertow-pull-request:

ctomc:
this is ok to test

PR #730, opened 2019-03-01
Fix Javadoc errors

tofflos:
This fixes all Javadoc **errors** by replacing all self-closing elements and then correcting the remaining five errors on a case-by-case basis. Javadoc generation will now succeed with the `doclint all` setting but I have left it at `doclint none` to prevent the build from being flooded by Javadoc **warnings**.


undertow-pull-request:

PR #735, opened 2019-03-20
Normalize file path before checking if file exists

brostbeef:
Reason: Fix "false" file doesn't exist issues

Trace logs from my web application:
2019-03-19 12:39:50,954 [issuer: Test Account] TRACE [io.undertow.server.handlers.resource.PathResourceManager] (default task-29) Failed to get path resource CustomerInfo/../WEB-INF/jsp/customer/profile/IndividualProfileQuery.jsp from path resource manager with base /appl/jboss/jboss-eap-7.2-AFStagJ/standalone/tmp/vfs/temp/temp8235cca585546399/content-ab1ec3bf15aee59c/, as the path did not exist

However, same server and application running on Windows:
12:50:48,418 TRACE [io.undertow.server.handlers.resource.PathResourceManager] (default task-11) Found path resource CustomerInfo/../WEB-INF/jsp/customer/profile/IndividualProfileQuery.jsp from path resource manager with base C:\Develop\jboss-eap-7.2-AFStagJ\standalone\tmp\vfs\temp\temp9584729df25c03a9\content-78189cb1518d12e\
12:50:51,224 TRACE [io.undertow.server.handlers.resource.PathResourceManager] (default task-11) Found path resource WEB-INF/jsp/customer/profile/IndividualProfileQuery.jsp from path resource manager with base C:\Develop\jboss-eap-7.2-AFStagJ\standalone\tmp\vfs\temp\temp9584729df25c03a9\content-78189cb1518d12e\

"/CustomerInfo" is a struts module which gets added to the forward path.  This is NOT a real directory.  In JBoss EAP 6.2 running on Java 7u67, this pathing does not create an issue.  JBoss EAP 7.2 running on Java 8u181 does create an issue.

For some reason, Files.exists(file) on windows has no issue, but in Linux, there is an issue.  Normalizing the path first fixes it.

My questions if my proposed change isn't accepted are: 
Why wouldn't you normalize the Path?
What's the point of walking the directory structure needlessly?

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/resource/PathResourceManager.java
+++ b/core/src/main/java/io/undertow/server/handlers/resource/PathResourceManager.java
@@ -194,7 +194,7 @@ public class PathResourceManager implements ResourceManager  {
             path = p;
         }
         try {
-            Path file = fileSystem.getPath(base, path);
+            Path file = fileSystem.getPath(base, path).normalize();
             String normalizedFile = file.normalize().toString();
             if(!normalizedFile.startsWith(base)) {
                 if(normalizedFile.length() == base.length() - 1) {


undertow-pull-request:

core/src/main/java/io/undertow/server/handlers/resource/PathResourceManager.java
```diff
@@ -194,7 +194,7 @@ public Resource getResource(final String p) {
             path = p;
         }
         try {
-            Path file = fileSystem.getPath(base, path);
+            Path file = fileSystem.getPath(base, path).normalize();
             String normalizedFile = file.normalize().toString();
```
stuartwdouglas: There should only be one call to normalize(), but in general we need to be really careful here as there are security issues around symlinks. In your failing example is there a symlink somewhere on the path?

brostbeef: My failing does not have any symlinks on the path.  Quite simply the resource gets requested with a false directory.  It has to do with the inner workings of struts 1.

Any thoughts as to what's changed between the undertow subsystem in JBoss EAP 7.2 and the old subsystem in JBoss EAP 6.2?  Although I would agree that the pathing is dumb, it's in a very large scale application that would require hundreds of changes (or ~70 empty directories created at the root of the WAR file) in order to work using the new system based on my tests.

Any/All help is greatly appreciated :-)

stuartwdouglas:
There is a simple workaround, you could create a ServletExtension that wraps the ResourceManager implementation and calls io.undertow.util.CanonicalPathUtils#canonicalize() on the path before delegating to the underlying implementation.

I don't know if I really want to support this out of the box, it would need a lot of thought about any potential security implications.

brostbeef:
I will give your suggestion a try... thank you!

PR #749, opened 2019-04-24
ReceiverTestCase#testAsyncReceiveWholeBytesFailed[proxy][http2] fails…

rhusar:
… intermittently on Windows

At this point (40% failure rate) this is a good candidate for disabling.. opened https://issues.jboss.org/browse/UNDERTOW-1531.

### DIFF
--- a/core/src/test/java/io/undertow/server/handlers/ReceiverTestCase.java
+++ b/core/src/test/java/io/undertow/server/handlers/ReceiverTestCase.java
@@ -174,6 +174,9 @@ public class ReceiverTestCase {
 
     @Test
     public void testAsyncReceiveWholeBytesFailed() throws Exception {
+        // TODO un-@Ignore - https://issues.jboss.org/browse/UNDERTOW-1531 - ReceiverTestCase#testAsyncReceiveWholeBytesFailed[proxy][http2] fails intermittently on Windows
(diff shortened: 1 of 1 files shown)


core/src/test/java/io/undertow/server/handlers/ReceiverTestCase.java
```diff
@@ -174,6 +174,9 @@ public void testAsyncReceiveWholeBytes() {
 
     @Test
     public void testAsyncReceiveWholeBytesFailed() throws Exception {
+        // TODO un-@Ignore - https://issues.jboss.org/browse/UNDERTOW-1531 - ReceiverTestCase#testAsyncReceiveWholeBytesFailed[proxy][http2] fails intermittently on Windows
+        org.junit.Assume.assumeFalse(System.getProperty("os.name").startsWith("Windows"));
```
carterkozak: I think this ignores the http/1.x variants as well, is that intentional?

May want to use something along these lines:
```java
assumeFalse(DefaultServer.isH2() && SystemUtils.IS_OS_WINDOWS);
````

rhusar: @carterkozak You are right, not intentional just too hasty :) it's really just the http2 variant that fails frequently we should ignore.

PR #750, opened 2019-04-26
Fix type of Setter for close listener

olegshtch:


undertow-pull-request:

stuartwdouglas:
this is ok to test

fl4via:
@olegshtch , I created a Jira to evaluate the impact of this type of change here: https://issues.redhat.com/browse/UNDERTOW-2631
if we decide to follow this route, it makes sense to change all getCloseSetters and other methods as such.
I am closing this PR for now, as it will require more investigation and rework if we follow that route.

PR #752, opened 2019-05-02
 UNDERTOW-1534 HTTP/2 can enter an infinite loop

stuartwdouglas:
Also re-enables the sender test case to see if this is the root cause


gaol:
Are the test failures related? It failed again after I triggered.

fl4via:
Retest this please

gaol:
https://issues.jboss.org/browse/JBEAP-16855 got verified using this patch, so it looks good.

uherberg:
I can't see the detailed description of this error (some permission issue), but we also see an infinite loop. I am hoping this will fix that.

stuartwdouglas:
Retest this please

stuartwdouglas:
Retest this please

PR #754, opened 2019-05-03
UNDERTOW-1533 Send SNI when connecting via HTTP Proxy

criege:
This is basically the same code as is used in `UndertowXnioSsl.StreamConnectionChannelListener#handleEvent` to enforce SNI on the connection.


undertow-pull-request:

stuartwdouglas:
This is ok to test

PR #755, opened 2019-05-08
[UNDERTOW-1525] At BlockingWriterSenderImpl, move writer.checkError t…

fl4via:
…o close

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/core/BlockingWriterSenderImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/core/BlockingWriterSenderImpl.java
@@ -100,11 +100,7 @@ public class BlockingWriterSenderImpl implements Sender {
         }
         writer.write(data);
 
-        if (writer.checkError()) {
-            callback.onException(exchange, this, new IOException());
-        } else {
-            invokeOnComplete(callback);
-        }
+        invokeOnComplete(callback);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


servlet/src/main/java/io/undertow/servlet/core/BlockingWriterSenderImpl.java
```diff
@@ -183,14 +175,19 @@ private void performTransfer(FileChannel source, IoCallback callback) {
...
+            invokeOnComplete(callback);
+        }
     }
 
     @Override
     public void close() {
         if(exchange.getAttachment(ServletRequestContext.ATTACHMENT_KEY).getDispatcherType() != DispatcherType.INCLUDE) {
             IoUtils.safeClose(writer);
         }
+        writer.checkError();
```
fl4via: @stuartwdouglas does it make sense to you to invoke checkError from close without the callback? Please, let me know what you think of this fix.

stuartwdouglas: Why are you calling it without checking the return value?

fl4via: @stuartwdouglas  That's the point. If I check the return value, there's no point, because there's no callback? Should I throw a RuntimeException? This is the part I'm lost at.

stuartwdouglas:
Retest this please

PR #756, opened 2019-05-10
UNDERTOW-1536: fixed client auth when using proxy protocol

uherberg:
When Proxy Protocol (both version v1 and v2) is enabled, client authentication doesn't work. During the TLS Handshake, the server hello doesn't include the Certificate Request, so the client never sends its certificates. With the same configuration but disabled proxy protocol, the server hello is correct (and the client sends its certs). I verified this running a local HAProxy with enabled proxy protocol. This fix sets the SSLEngine configuration to properly request the client certificates.

### DIFF
--- a/core/src/main/java/io/undertow/protocols/ssl/UndertowXnioSsl.java
+++ b/core/src/main/java/io/undertow/protocols/ssl/UndertowXnioSsl.java
@@ -21,6 +21,8 @@ package io.undertow.protocols.ssl;
 import static org.xnio.IoUtils.safeClose;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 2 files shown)


uherberg:
Not sure why the Windows tests failed. Tests ran successfully on my local host. Linux tests pass.

fl4via:
Retest this please

uherberg:
Good catch. I’ll fix that

uherberg:
@fl4via I fixed it, and also copied some code to set the cipher order (copied from UndertowAcceptingSslChannel).

fl4via:
Retest this please

uherberg:
@fl4via I squashed the two commits into a single one, as you suggested.

uherberg:
@fl4via Any update when this can be merged and when Undertow 2.0.21 can be released? We can't use Undertow until this is fixed, since our services require mTLS.

PR #757, opened 2019-05-15
UNDERTOW-1539 Allow checking server DNS name vs. its certificate

criege:
The patch is pretty self-explanatory. I'm a bit bugged by the fact that the `SSLEngine` itself is created by different means in two places but that's how it is and I don't think changing this right now would provide any more benefits.


fl4via:
This is ok to test

fl4via:
Retest this please

fl4via:
Retest this please

fl4via:
Retest this please

PR #760, opened 2019-05-22
UNDERTOW-1543: avoid create too many random instances ...

wenhoujx:
perf testing websocket in my application, flight recorder show this class is generating 78mb `random` instances, which could be just be one b/c i don't think it's using random state.

image

### DIFF
--- a/core/src/main/java/io/undertow/websockets/core/protocol/version07/WebSocket07FrameSinkChannel.java
+++ b/core/src/main/java/io/undertow/websockets/core/protocol/version07/WebSocket07FrameSinkChannel.java
@@ -29,7 +29,7 @@ import io.undertow.connector.PooledByteBuffer;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * {@link StreamSinkFrameChannel} implementation for writing WebSocket Frames on {@link io.undertow.websockets.core.WebSocketVersion#V08} connections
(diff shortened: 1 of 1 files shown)


undertow-pull-request:

core/src/main/java/io/undertow/websockets/core/protocol/version07/WebSocket07FrameSinkChannel.java
```diff
@@ -41,7 +41,7 @@
     private final Masker masker;
     private volatile boolean dataWritten = false;
     protected final ExtensionFunction extensionFunction;
-    private final Random random = new Random();
+    private static final Random random = new Random();
```
carterkozak: Would we be better off replacing the one use of this random instance with `ThreadLocalRandom.current().nextInt()`?

carterkozak: From [Random javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html):

> Instances of java.util.Random are threadsafe. However, the concurrent use of the same java.util.Random instance across threads may encounter contention and consequent poor performance. Consider instead using ThreadLocalRandom in multithreaded designs.

wenhoujx: y, that's faster, let me update the pr.

stuartwdouglas:
This is ok to test

PR #761, opened 2019-05-23
UNDERTOW-1543: SimpleNonceManager uses ThreadLocalRandom

carterkozak:
Reduces potential contention between threads using a shared
Random instance.


carterkozak:
I've piggy-backed on UNDERTOW-1543, I can file another ticket for this if you like. This one is more cleanup than anything.

fl4via:
@carterkozak  that's alright, no need to create another Jira, thanks btw!

PR #762, opened 2019-05-23
Remove unused HttpString.hashCodeBase

carterkozak:
This field was previously used in an alternative murmur-3 hash
implementation, but was not deleted with the code it supported
in 15d3d02efc3c54216d189b7df23d77fc86112cd4.


fl4via:
This is ok to test

PR #743, opened 2019-04-16
Add an `undertow-benchmarks` module with basic JMH benchmarks

carterkozak:
Currently this only provides four simple benchmarks for blocking I/O
functionality, but may be expanded to benchmark additional features.

PR #763, opened 2019-05-23
Exclude JMH generated packages from spotbugs

carterkozak:


carterkozak:
I didn't realize that spotbugs was failing on #743, this should resolve it.

PR #772, opened 2019-06-03
UNDERTOW-1550 Make undertow build fine on JDK11

jaikiran:
The commit here fixes the issue reported in https://issues.jboss.org/browse/UNDERTOW-1550 letting the `jboss-parent` pom dictate the version of `maven-bundle-plugin`. The `jboss-parent` pom version `35` correctly uses a version of `maven-bundle-plugin` which understands Java 9+ multi-release jars and allows builds to pass on JDK11.

P.S: With this change, the build works fine with both JDK 8 and 11.


fl4via:
This is ok to test

PR #773, opened 2019-06-03
[3.x] Add an `undertow-benchmarks` module with basic JMH benchmarks

carterkozak:
Currently this only provides four simple benchmarks for blocking I/O
functionality, but may be expanded to benchmark additional features.

Note that this benchmark fails against 3.x where it exposes a race condition resulting in `NoHttpResponseException: The target server failed to respond`. Reducing the thread count from `32` to `1` allows the benchmark to succeed.


fl4via:
I'm working on the automatic tests for validating 3.x PRs, as soon as this is done I'll merge this. Thanks for your PR!

fl4via:
I'll be merging this regardless of having fixed the tests. We will revisit the 3.x topic later this year, stay tuned!

PR #774, opened 2019-06-03
UNDERTOW-1553 Use the correct character encoding for the reader/writer

jaikiran:
The commit here fixes the issue noted in this forum thread https://developer.jboss.org/message/989396#989396 (and tracked in JIRA https://issues.jboss.org/browse/UNDERTOW-1553). 

This commit makes sure the reader's charset encoding takes into account the encoding set on servletcontext level and then falls back to the default container level encoding. The commit does a similar thing with the response's writer.

The changes in this commit follow the rules noted in the javadoc of `javax.servlet.ServletRequest#getCharacterEncoding` and `javax.servlet.ServletResponse#getCharacterEncoding`

The existing testcase has been updated to reproduce and verify this fix.

### DIFF
--- a/core/src/test/java/io/undertow/testutils/HttpClientUtils.java
+++ b/core/src/test/java/io/undertow/testutils/HttpClientUtils.java
@@ -24,6 +24,7 @@ import org.apache.http.HttpResponse;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
 /**
@@ -36,11 +37,15 @@ public class HttpClientUtils {
     }
 
     public static String readResponse(final HttpResponse response) throws IOException {
+        return readResponse(response, StandardCharsets.UTF_8);
+    }
+
+    public static String readResponse(final HttpResponse response, final Charset charset) throws IOException {
         HttpEntity entity = response.getEntity();
         if(entity == null) {
             return "";
         }
-        return readResponse(entity.getContent());
+        return readResponse(entity.getContent(), charset);
     }
 
     public static String readResponse(InputStream stream) throws IOException {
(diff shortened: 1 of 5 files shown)


jaikiran:
The Windows failures are unrelated to this change and those failures appear to be related to the environment.
However, the linux one has uncovered an issue with one part of this change. The linux build has failed with:

```[06:26:21]
[Step 1/1] [ERROR] Found reliance on default encoding in io.undertow.servlet.spec.HttpServletRequestImpl.getReader(): new java.io.InputStreamReader(InputStream) [io.undertow.servlet.spec.HttpServletRequestImpl] At HttpServletRequestImpl.java:[line 879] DM_DEFAULT_ENCODING
```
This error complains about this line, in the commit, https://github.com/undertow-io/undertow/pull/774/commits/0776096a3624f1fb79728d3e5ce7cd5665660288#diff-64684efd4a45ac1559b02fc2b6657e09R879

```reader = new BufferedReader(charSet == null ? new InputStreamReader(exchange.getInputStream())
                    : new InputStreamReader(exchange.getInputStream(), charSet));```
Previously, that line used to be:
```reader = new BufferedReader(new InputStreamReader(exchange.getInputStream(), charSet));```
I changed it because the `charSet` being passed could be potentially (in theory) be null and that would result in a NPE from the constructor of `InputStreamReader`. So I decided to prevent that and use the other constructor instead, which will end up using the Java system default encoding, which is what the build failure is complaining about.
I think a better version of my change perhaps could be:

```reader = new BufferedReader(charSet == null ? new InputStreamReader(exchange.getInputStream(), StandardCharsets.ISO_8859_1)
                    : new InputStreamReader(exchange.getInputStream(), charSet));```
i.e. default to ```StandardCharsets.ISO_8859_1``` if the charset cannot be determined (because of it being not configured at any level). I'll update this PR accordingly in a while.

jaikiran:
> I think a better version of my change perhaps could be:
> 
> reader = new BufferedReader(charSet == null ? new InputStreamReader(exchange.getInputStream(), StandardCharsets.ISO_8859_1)
>                     : new InputStreamReader(exchange.getInputStream(), charSet));
> 
> i.e. default to StandardCharsets.ISO_8859_1 if the charset cannot be determined (because of it being not configured at any level). I'll update this PR accordingly in a while.

PR has been updated

PR #779, opened 2019-06-13
[UNDERTOW-1559] At ApplicationListeners.requestInitialized, notify al…

fl4via:
…l "requestInitialized" notified listeners of requestDestroyed so they don't end in an inconsistent state


fl4via:
Jira: https://issues.jboss.org/browse/UNDERTOW-1559

fl4via:
This is ok to test

PR #781, opened 2019-06-14
[UNDERTOW-1551] Add new exchange attributes for request cookie and re…

msfm:
…sponse cookie

https://issues.jboss.org/browse/UNDERTOW-1551


fl4via:
This is ok to test

PR #784, opened 2019-06-24
Use proper peer address for proxied connections

criege:
This fixes UNDERTOW-1564 by using the host name of the indicated destination URI rather than the address of the proxy connection.


criege:
> Please, add Jira number UNDERTOW-1564 to the commit message

Done.

PR #785, opened 2019-06-25
UNDERTOW-1565 Don't encode query strings and URI paths while redirecting

jaikiran:
The commit here fixes the issue reported in https://issues.jboss.org/browse/UNDERTOW-1565

The code in `SinglePortConfidentialityHandler` which deals with forming the redirect URI was using the 7 argument constructor of `java.net.URI`, which as per its javadoc, internally encodes the passed parameters, like the `queryString`. We do not want that to happen, since the query string and other components of the URI that we are building aren't expected to be decoded/encoded, by the container, while we are building the redirect URI.

The commit here uses a `StringBuilder` to form the redirect URI and then passes it to the single argument constructor of `java.net.URI` which doesn't do the encoding of the passed parameter, unlike the 7 argument constructor.

The commit also includes an update to an existing testcase to reproduce the issue and verify the fix.

### DIFF
--- a/core/src/main/java/io/undertow/security/handlers/SinglePortConfidentialityHandler.java
+++ b/core/src/main/java/io/undertow/security/handlers/SinglePortConfidentialityHandler.java
@@ -43,10 +43,12 @@ public class SinglePortConfidentialityHandler extends AbstractConfidentialityHan
         return getRedirectURI(exchange, redirectPort);
     }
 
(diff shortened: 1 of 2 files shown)


jaikiran:
retest this please

jaikiran:
These failures don't look related to this change (and in fact the tests work fine locally)

PR #786, opened 2019-06-25
UNDERTOW-1567 Redirect to absolute URL with special characters broken

TomasHofman:
https://issues.jboss.org/browse/UNDERTOW-1567
https://issues.jboss.org/browse/JBEAP-16826

Use regex to determine URI scheme instead of `new URI(location).getScheme()`.

The presence of scheme in given `location` is used to determine whether `location` is an absolute or relative URI. The URI instantiation fails when `location` contains technically illegal characters like {}, the `location` is then always determined as relative. This was not the case in previous EAP versions.

### DIFF
--- a/core/src/main/java/io/undertow/util/URLUtils.java
+++ b/core/src/main/java/io/undertow/util/URLUtils.java
@@ -19,8 +19,7 @@
 package io.undertow.util;
 
 import java.io.UnsupportedEncodingException;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


TomasHofman:
Retest this please.

xstefank:
retest this please

TomasHofman:
Failures look unrelated to me. I opened PR https://github.com/undertow-io/undertow/pull/791 to fix `ProxyProtocolTestCase.testProxyProtocolV2Local` one.

fl4via:
Thanks for looking into the test failures @TomasHofman ! It is definitely on my list to get rid of those

PR #787, opened 2019-06-28
[UNDERTOW-1569] HttpServletRequest getLocalName() returns IP instead …

tmiyargi:
…of hostname

Issue: https://issues.jboss.org/browse/UNDERTOW-1569

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
@@ -981,9 +981,16 @@ public final class HttpServletRequestImpl implements HttpServletRequest {
         return exchange.getSourceAddress().getPort();
     }
 
+    /**
+     * String java.net.InetAddress.getHostName()
+     * Gets the host name for this IP address.
+     * If this InetAddress was created with a host name, this host name will be remembered and returned; otherwise, a reverse name lookup will be performed and the result will be returned based on the system configured name lookup service. If a lookup of the name service is required, call getCanonicalHostName.
+     * If there is a security manager, its checkConnect method is first called with the hostname and -1 as its arguments to see if the operation is allowed. If the operation is not allowed, it will return the textual representation of the IP address.
+     * @see InetAddres#getHostName
+     */
     @Override
     public String getLocalName() {
-        return exchange.getDestinationAddress().getHostString();
+        return exchange.getDestinationAddress().getHostName();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
```diff
@@ -983,7 +983,12 @@ public int getRemotePort() {
 
     @Override
     public String getLocalName() {
-        return exchange.getDestinationAddress().getHostString();
+        String hostName = exchange.getDestinationAddress().getHostName();
+        if (hostName == null) {
```
jaikiran: Hi @tmiyargi,

Looking at the javadoc of `ServletRequest#getLocalName()`, this method should never return a IP address and instead should be a host name. So I don't think we should return an IP address if we can't resolve the host name.

tmiyargi: @jaikiran you are correct, but what about backward compatibility? It has been working like this for a long time

jaikiran: Hi @tmiyargi 
@fl4via and/or @stuartwdouglas will have a final say on this one. I'm just a lurker in this project.
Having said that, I think the backward compatiblity of this shouldn't play a role here, because this bug fix is about fixing a wrong value that was being retuned, before this change.

tmiyargi: @jaikiran I've found in java doc of inetAddr the following:
`/**
     * Gets the host name for this IP address.
     *
     * <p>If this InetAddress was created with a host name,
     * this host name will be remembered and returned;
     * otherwise, a reverse name lookup will be performed
     * and the result will be returned based on the system
     * configured name lookup service. If a lookup of the name service
     * is required, call
     * {@link #getCanonicalHostName() getCanonicalHostName}.
     *
     * <p>If there is a security manager, its
     * {@code checkConnect} method is first called
     * with the hostname and {@code -1}
     * as its arguments to see if the operation is allowed.
     * If the operation is not allowed, it will return
     * the textual representation of the IP address.
     *
     * @return  the host name for this IP address, or if the operation
     *    is not allowed by the security check, the textual
     *    representation of the IP address.
     *
     * @see InetAddress#getCanonicalHostName
     * @see SecurityManager#checkConnect
     */`

That means an ip could be returned, I don't think there is another way @fl4via, what do you think?

xstefank:
@fl4via review ping. Can you take a look, please?

fl4via:
@xstefank yes, this will go in the next tag

servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
fl4via: @tmiyargi I agree, I'm approving this

PR #793, opened 2019-07-23
HttpServletRequest.getRequestedSessionID() is incorrectly returning a…

tmiyargi:
… newly generated session ID instead of the requested ID in EAP 7 when using URL session tracking

Issue: https://issues.jboss.org/browse/UNDERTOW-1575

This is a proposal for fixing the issue above, it makes a test fail (right now set to Ignore) because it was testing the opposite, when looking at java doc SessionManager.createSession(true), also testing this functionality in Tomcat behaves as this proposed fix.


xstefank:
@favila this is a blocker for CP, can you take a look, please?

xstefank:
@fl4via this is a blocker for CP, can you take a look, please?

PR #794, opened 2019-07-24
[UNDERTOW-1578] 401 Unauthorized should be returned when requesting a protected directory without trailing slash

gaol:
Issue: https://issues.jboss.org/browse/UNDERTOW-1578

Created a `RedirectDirHandler` to handle the redirect after security check.


gaol:
@stuartwdouglas @fl4via Could you please review? Thanks :)

stuartwdouglas:
LGTM

mibo:
Hi @fl4via is it planned to fix this in the `1.4.x` versions as well?
Regards, Michael

PR #798, opened 2019-08-04
UNDERTOW-1582 Fix NPE while setting up temporary directory

jaikiran:
The commit here fixes a NPE that's reported in https://issues.jboss.org/browse/UNDERTOW-1582. The change here uses the system level temporary directory as the base directory, if no explicit temporary directory is setup.

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/api/SecurityActions.java
+++ b/servlet/src/main/java/io/undertow/servlet/api/SecurityActions.java
@@ -1,0 +1,42 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package io.undertow.servlet.api;
+
+
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+final class SecurityActions {
+
+    private SecurityActions() {
+        // forbidden inheritance
+    }
+
+    static String getSystemProperty(final String prop) {
+        if (System.getSecurityManager() == null) {
+            return System.getProperty(prop);
+        } else {
+            return (String) AccessController.doPrivileged(new PrivilegedAction<Object>() {
+                public Object run() {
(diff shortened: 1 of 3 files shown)


servlet/src/main/java/io/undertow/servlet/core/ManagedServlet.java
```diff
@@ -104,7 +104,9 @@ public void setupMultipart(ServletContextImpl servletContext) {
                 if(locFile.isAbsolute()) {
                     tempDir = locFile;
                 } else {
-                    tempDir = servletContext.getDeployment().getDeploymentInfo().getTempPath().resolve(location);
+                    final Path baseTmpDir = servletContext.getDeployment().getDeploymentInfo().getTempPath();
+                    tempDir = baseTmpDir == null ? Paths.get(System.getProperty("java.io.tmpdir"), location)
+                            : baseTmpDir.resolve(location);
```
carterkozak: DeploymentManagerImpl uses `SecurityActions.getSystemProperty("java.io.tmpdir")`, I think we should use that utility here as well.

This issue has come up a few times in several different places, I wonder if we should initialize `DeploymentInfo.tempDir` with default value `SecurityActions.getSystemProperty("java.io.tmpdir")` instead of spot-fixing each time it's used elsewhere?

jaikiran: Sorry, I have been away, so couldn't reply earlier.
>> DeploymentManagerImpl uses SecurityActions.getSystemProperty("java.io.tmpdir"), I think we should use that utility here as well.

You are right, that needs to be done here too. I'll update accordingly.

>>This issue has come up a few times in several different places, I wonder if we should initialize DeploymentInfo.tempDir with default value SecurityActions.getSystemProperty("java.io.tmpdir") instead of spot-fixing each time it's used elsewhere?

The thing is, the DeploymenInfo has a setter which allows null value to be set to this member, which means that setting a default value won't solve this issue fully (some code could actually set a null value to overwrite this default value). I have an alternate solution in mind, which I will include in the update to this PR. Thank you @carterkozak for this review.

jaikiran:
Hello @carterkozak, @fl4via,
I've updated this PR to do a couple of things:
1. Use the privileged action if security manager is set, to look up the system property.
2. Introduce a new `requireTempPath` method on `DeploymentInfo` which returns a non-null value for temp directory, by falling back to the system level temporary directory, if no explicit temp directory is set on the `DeploymentInfo`. This way, the callers don't have to keep doing null checks all over the place. I haven't changed any existing uses of getTempDir and instead they can be changed/updated if/when relevant.

servlet/src/main/java/io/undertow/servlet/api/SecurityActions.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
fl4via: Please, update the year

jaikiran:
Hi @fl4via, I just updated this PR to fix the copyright year. Thank you for the review.

fl4via:
@jaikiran thank you for the PR!

PR #799, opened 2019-08-04
UNDERTOW-1581: Fix RoutingHandler allMethodsMatcher validation

carterkozak:
Updates the allMethodsMatcher to lookup templates using the
match method instead of get in order to normalize the input.

Previously registration would throw an exception when routes
were added for the same path template with different parameter names
despite being equivalent.

### DIFF
--- a/core/src/test/java/io/undertow/server/handlers/RoutingHandlerTestCase.java
+++ b/core/src/test/java/io/undertow/server/handlers/RoutingHandlerTestCase.java
@@ -126,6 +126,12 @@ public class RoutingHandlerTestCase {
                         exchange.getResponseSender().send("posted foo");
                     }
                 })
+                .add(Methods.POST, "/foo/{baz}", new HttpHandler() {
+                    @Override
+                    public void handleRequest(HttpServerExchange exchange) throws Exception {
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/undertow/server/handlers/RoutingHandlerTestCase.java
```diff
@@ -126,6 +126,12 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
                         exchange.getResponseSender().send("posted foo");
                     }
                 })
+                .add(Methods.POST, "/foo/{baz}", new HttpHandler() {
+                    @Override
+                    public void handleRequest(HttpServerExchange exchange) throws Exception {
+                        exchange.getResponseSender().send("foo-path" + exchange.getQueryParameters().get("bar"));
+                    }
+                })
```
carterkozak: Without this fix, adding this handler causes all tests to fail.

fl4via:
This PR introduce a regression in the Windows testsuite:
HttpContinueAcceptingHandlerTestCase.testHttpContinueRejected
I'm running once more just to be safe, but this PR probably will need some investigation to find out why we are seeing that failure

PR #801, opened 2019-08-07
Expose the ServerConnection created timestamp

eagleshine:


undertow-pull-request:

fl4via:
This is ok to test

fl4via:
Hi @eagleshine ! Thanks for your PR!
Undertow is not compiling, the following error occurs: Compilation failure /store/work/tc-work/b455211ade1ff869/core/src/main/java/io/undertow/server/protocol/http2/Http2ServerConnection.java:[83,8] io.undertow.server.protocol.http2.Http2ServerConnection is not abstract and does not override abstract method getCreatedTimestamp() in io.undertow.server.ServerConnection
Can you fix it?
Also, what is the drive behing the timestamp exposition? Are you using that for some logging or something?

eagleshine:
We have an HTTPS server that has some logic to decide when to return the header "Connection: close" to the client using the connection created timestamp. We are using some internally developed web framework but want to migrate to either Jetty or Undertow.

core/src/main/java/io/undertow/server/protocol/http2/Http2ServerConnection.java
```diff
@@ -110,6 +111,7 @@ public Http2ServerConnection(Http2Channel channel, Http2StreamSourceChannel requ
         originalSourceConduit = new StreamSourceChannelWrappingConduit(requestChannel);
         this.conduitStreamSinkChannel = new ConduitStreamSinkChannel(responseChannel, originalSinkConduit);
         this.conduitStreamSourceChannel = new ConduitStreamSourceChannel(channel, originalSourceConduit);
+        this.createdTimestamp = System.currentTimeMillis();
```
carterkozak: Is this correct? If memory serves, a new `Http2ServerConnection` object is created for each request, so the createdTimestamp will be set at the same time as `HttpServerExchange.requestStartTime` (which uses the high resolution nanoTime, and is default disabled for performance).

eagleshine: OK. Looks like the semantics of the Connection changed in Http2: the equivalent to that of Http in Http2 is `Http2Channel`. And you're right that a new `Http2ServerConnection` object is created for each request, however, this is still the connection created timestamp, just not useful anymore.

fl4via: @eagleshine does that mean that this PR won't work the way you intended it to? If it is no longer valid, please close it. If that's not the case, please let me know

fl4via:
Closing as we got no response from PR author for a long time. If you still are interested in this change, please let us know.

PR #807, opened 2019-08-30
UNDERTOW-1581: Fix RoutingHandler allMethodsMatcher validation

carterkozak:
Updates the allMethodsMatcher to lookup templates using the
match method instead of get in order to normalize the input.

Previously registration would throw an exception when routes
were added for the same path template with different parameter names
despite being equivalent.


fl4via:
This is ok to test

fl4via:
I'm closing this as it is a duplicate @carterkozak created at my request for more CI checking.

PR #808, opened 2019-09-13
Fixing rewrite handler so that it properly routes to destination servlet

kasinskim:
## Description
- Using `RewriteHandler` does not cause requests to be routed to destination servlet, requests are routed to original servlets instead

  With the rule defined as `RewriteRule /foo1 /bar1` request were still send to `/foo1` servlet

- Enhancing test case to replicate rewrite scenario by removing original servlet that could handle inappropriately rewritten request

## Fix
Setting `requestURI` on `HttpServerExchange` solves the problem


undertow-pull-request:

fl4via:
This is ok to test

PR #813, opened 2019-10-04
[do not merge] XNIO-350 reproducer

carterkozak:
It's more difficult to reproduce here than on my internal
configuration, I can iterate on making it easier to surface.


benchmarks/src/main/java/io/undertow/benchmarks/SimpleBenchmarkState.java
```diff
@@ -44,11 +45,14 @@
  */
 @State(Scope.Benchmark)
 public class SimpleBenchmarkState {
+    static {
+        System.setProperty("xnio.nio.alt-queued-server", "true");
```
carterkozak: not strictly necessary because this is the default.

PR #816, opened 2019-10-10
UNDERTOW-1600 Enhance SameSite cookie support

msfm:
https://issues.jboss.org/browse/UNDERTOW-1600

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/CookieImpl.java
+++ b/core/src/main/java/io/undertow/server/handlers/CookieImpl.java
@@ -18,9 +18,10 @@
 
 package io.undertow.server.handlers;
 
+import java.util.Arrays;
 import java.util.Date;
-import java.util.Locale;
 
+import io.undertow.UndertowLogger;
 import io.undertow.UndertowMessages;
 
 /**
@@ -41,7 +42,6 @@ public class CookieImpl implements Cookie {
     private String comment;
     private boolean sameSite;
     private String sameSiteMode;
-
 
     public CookieImpl(final String name, final String value) {
         this.name = name;
@@ -163,20 +163,14 @@ public class CookieImpl implements Cookie {
     }
 
     @Override
-    public Cookie setSameSiteMode(final String sameSiteMode) {
-        if (sameSiteMode != null) {
-            switch (sameSiteMode.toLowerCase(Locale.ENGLISH)) {
-                case "strict":
-                    this.setSameSite(true);
-                    this.sameSiteMode = "Strict";
-                    break;
-                case "lax":
-                    this.setSameSite(true);
-                    this.sameSiteMode = "Lax";
-                    break;
-                default:
-                    throw UndertowMessages.MESSAGES.invalidSameSiteMode(sameSiteMode);
-            }
+    public Cookie setSameSiteMode(final String mode) {
+        final String m = CookieSameSiteMode.lookupModeString(mode);
+        if (m != null) {
+            UndertowLogger.REQUEST_LOGGER.tracef("Setting SameSite mode to [%s] for cookie [%s]", m, this.name);
+            this.sameSiteMode = m;
+            this.setSameSite(true);
+        } else {
+            UndertowLogger.REQUEST_LOGGER.warnf(UndertowMessages.MESSAGES.invalidSameSiteMode(mode, Arrays.toString(CookieSameSiteMode.values())), "Ignoring specified SameSite mode [%s] for cookie [%s]", mode, this.name);
         }
         return this;
     }
--- a/core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
@@ -41,24 +41,21 @@ public class SameSiteCookieHandler implements HttpHandler {
     private final String mode;
     private final Pattern cookiePattern;
     private final boolean enableClientChecker;
+    private final boolean addSecureForNone;
 
     public SameSiteCookieHandler(final HttpHandler next, final String mode) {
-        this(next, mode, null, true, true);
-    }
-
-    public SameSiteCookieHandler(final HttpHandler next, final String mode, final boolean enableClientChecker) {
-        this(next, mode, null, true, enableClientChecker);
+        this(next, mode, null, true, true, true);
     }
 
     public SameSiteCookieHandler(final HttpHandler next, final String mode, final String cookiePattern) {
-        this(next, mode, cookiePattern, true, true);
+        this(next, mode, cookiePattern, true, true, true);
     }
 
     public SameSiteCookieHandler(final HttpHandler next, final String mode, final String cookiePattern, final boolean caseSensitive) {
-        this(next, mode, cookiePattern, caseSensitive, true);
+        this(next, mode, cookiePattern, caseSensitive, true, true);
     }
 
-    public SameSiteCookieHandler(final HttpHandler next, final String mode, final String cookiePattern, final boolean caseSensitive, final boolean enableClientChecker) {
+    public SameSiteCookieHandler(final HttpHandler next, final String mode, final String cookiePattern, final boolean caseSensitive, final boolean enableClientChecker, final boolean addSecureForNone) {
         this.next = next;
         this.mode = mode;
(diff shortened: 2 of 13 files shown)


msfm:
Sorry for the late response. I've just updated this PR with add a utility class SameSiteNoneIncompatibleClientChecker that can detect user agents which are incompatible with the SameSite=None attribute. I have also updated the description in the JIRA UNDERTOW-1600.

core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
```diff
@@ -0,0 +1,143 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
fl4via: we have different licensed headers in the files. This one has the wrong year.

core/src/main/java/io/undertow/util/SameSiteNoneIncompatibleClientChecker.java
```diff
@@ -0,0 +1,152 @@
+/*
```
fl4via: Ditto for license header

core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public SameSiteCookieHandler(final HttpHandler next, final String mode, final String cookieName) {
+        this(next, mode, cookieName, false);
+    }
+
+    public SameSiteCookieHandler(final HttpHandler next, final String mode, final String cookieName, final boolean disableClientChecker) {
+        this.next = next;
+        this.mode = mode;
+        this.cookieName = cookieName;
```
fl4via: It would be ideal with this cookie name were a pattern

core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private static final Logger log = Logger.getLogger(SameSiteCookieHandler.class);
+
+    private final HttpHandler next;
+    private final String mode;
+    private final String cookieName;
+    private final boolean disableClientChecker;
+
+    public SameSiteCookieHandler(final HttpHandler next, final String mode) {
+        this(next, mode, null, false);
```
fl4via: do we want to have a disable client checker with a default value set to false, or do we want to have a enable-client-checker with a default value set to true? I wonder how does this work in the long term pov, do we expect to still have a long list of incompatible user agents in the future?

core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
msfm: Fixed.

core/src/main/java/io/undertow/util/SameSiteNoneIncompatibleClientChecker.java
msfm: Fixed.

core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
msfm: Done. I've changed the parameter from "cookie-name" to "cookie-pattern" which can support regex. Also, I've added a new parameter "case-sensitive" (true by default) which can support incase-sensitive match by setting it to false.

core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
msfm: I have no preference whether disable-client-checker (false by default) or enable-client-checker (true by default). 

I'm not sure how long we should keep the client checker logic. Future user-agents will not be incompatible, but I guess some old incompatible user-agents (e.g. mobile browsers on Android or iOS) can remain for a longer time than we expect. So, I think we should enable the client checker by default.

fl4via: I wonder if this should be enabled by default. If the browser is old and does not support "None", it will fall back to ignoring the attribute completely in case it finds a SameSite=None. In the old specs, it says that anything different from Lax or Strick should cause the attribute to be ignored. Isn't this the same behavior we would get from a browser that supports SameSite="None" anyway?

fl4via: Please, ignore. I see that some browser versions were rejecting the cookie instead of ignoring it.

core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
```diff
@@ -0,0 +1,139 @@
...
+        this.disableClientChecker = disableClientChecker;
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        if (mode != null) {
+            exchange.addResponseCommitListener(new ResponseCommitListener() {
+                @Override
+                public void beforeCommit(HttpServerExchange exchange) {
+                    if (mode.equalsIgnoreCase(Cookie.SameSiteMode.None.toString())) {
```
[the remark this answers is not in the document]
msfm: > For performance reasons, we need to change mode to the enum

Hmm..., "mode" is not the enum type at this stage because `Cookie#setSameSiteMode(String mode)` interface just accepts String. (It was my fault when I made the initial implementation... I understand the method should be better to accept the enum type, but we can not change the method due to compatibility reasons.)

Also, as `Cookie#setSameSiteMode(String mode)` can accept the "mode" with incase-sensitive, I think it's better to make `SameSiteCookieHandler` accept the "mode" with incase-sensitive. Therefore, as a user can specify the "mode" with incase-sensitive, we can not simply use `SameSiteMode.valueOf(mode)` to get the enum type from the "mode" value.
(`SameSiteMode.valueOf("None")` can return the enum type, but `SameSiteMode.valueOf("none")` throws `IllegalArgumentException`.) 


> and we should also verify if disableClientChecker is not true, right here, short-circuiting the check for mode if not needed.

Got it. From this suggestion, I think of another idea to avoid invoking the mode string comparison in every request.
 
I flip the flag name from "disableClientChecker" to "enableClientChecker". Then, As the client checker is required only for "SameSite=None",  "enableClientChecker" is set to true only if the mode is specified to "None" (incase-sensitive) in the constructor. The mode string comparison does not happen every request but only once in the constructor.

core/src/main/java/io/undertow/server/handlers/CookieImpl.java
```diff
@@ -163,20 +164,14 @@ public String getSameSiteMode() {
...
-                    throw UndertowMessages.MESSAGES.invalidSameSiteMode(sameSiteMode);
-            }
+    public Cookie setSameSiteMode(final String mode) {
+        final String m = Cookies.lookupSameSiteMode(mode);
+        if (m != null) {
+            UndertowLogger.REQUEST_LOGGER.tracef("Setting SameSite mode to [%s] for cookie [%s]", m, this.name);
+            this.sameSiteMode = m;
+            this.setSameSite(true);
+        } else {
+            UndertowLogger.REQUEST_LOGGER.warnf(UndertowMessages.MESSAGES.invalidSameSiteMode(mode, Arrays.toString(Cookie.SameSiteMode.values())), "Ignoring specified SameSite mode [%s] for cookie [%s]", mode, this.name);
```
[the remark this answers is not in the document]
msfm: If just throwing the exception here, a request processing always fails with "500 Internal Server Error" when an invalid SameSite mode is specified. 
I think it's better to just ignore such an invalid SameSite mode (= wrong setting like a typo) and output the WARN log message instead of failing the request processing.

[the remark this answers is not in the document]
fl4via: Okay, I see your point

fl4via:
@msfm  it is weird that we are having the SameSite test case failing once in a while. I think it is unrelated, and I'm running again.

fl4via:
I can't reproduce the failures isolatedely in my machine, I'll going to run the whole testsuite to see what happens.

PR #822, opened 2019-11-04
[UNDERTOW-1612] Add support to multiple cookies with same name and pa…

fl4via:
…th, but different domains

Jira: https://issues.jboss.org/browse/UNDERTOW-1612


fl4via:
This is ok to test

PR #830, opened 2019-12-10
[2.1] UNDERTOW-1622: Implement blocking IO timeouts

carterkozak:
Added two new UndertowOptions:
- `BLOCKING_READ_TIMEOUT`: The amount of time in milliseconds a single
  blocking read operation may take before a timeout IOException is thrown.
  Unlike Options.READ_TIMEOUT this only applies to blocking operations
  which can be helpful to prevent the worker pool from becoming saturated
  when clients stop responding.
- `BLOCKING_WRITE_TIMEOUT`: The amount of time in milliseconds a single
  blocking write operation may take before a timeout IOException is
  thrown. Unlike Options.WRITE_TIMEOUT this only applies to blocking
  operations which can be helpful to prevent the worker pool from becoming
  saturated when clients stop responding.

A version of `xnio-api` containing the fix for [XNIO-356](https://issues.redhat.com/browse/XNIO-356) is required
for these options to work correctly, otherwise erronious timeouts
may be reported when channel.await{Readable,Writable} returns
spuriously.

This commit does not apply the write timeout to Channels.flushBlocking
invocations because that will require either a local utility class
to support flushBlocking with a timeout, or a backport of [XNIO-357](https://issues.redhat.com/browse/XNIO-357).

### DIFF
--- a/core/src/main/java/io/undertow/io/UndertowOutputStream.java
+++ b/core/src/main/java/io/undertow/io/UndertowOutputStream.java
@@ -22,8 +22,11 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
+import java.util.concurrent.TimeUnit;
 
 import io.undertow.UndertowMessages;
+import io.undertow.UndertowOptions;
+import io.undertow.server.BlockingChannels;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.util.Headers;
 import org.xnio.Buffers;
@@ -53,6 +56,7 @@ public class UndertowOutputStream extends OutputStream implements BufferWritable
     private int state;
     private long written;
     private final long contentLength;
+    private final int writeTimeoutMillis;
 
     private static final int FLAG_CLOSED = 1;
     private static final int FLAG_WRITE_STARTED = 1 << 1;
@@ -67,6 +71,8 @@ public class UndertowOutputStream extends OutputStream implements BufferWritable
     public UndertowOutputStream(HttpServerExchange exchange) {
         this.exchange = exchange;
         this.contentLength = exchange.getResponseContentLength();
+        this.writeTimeoutMillis = exchange.getConnection().getUndertowOptions()
+                .get(UndertowOptions.BLOCKING_WRITE_TIMEOUT, -1);
     }
 
 
@@ -161,7 +167,7 @@ public class UndertowOutputStream extends OutputStream implements BufferWritable
                             break;
                         }
                     }
-                    Channels.writeBlocking(channel, buffers, 0, bufferCount);
+                    writeBlocking(channel, buffers, 0, bufferCount);
                     while (bytesWritten < len) {
                         //ok, it did not fit, loop and loop and loop until it is done
                         bufferCount = 0;
@@ -182,7 +188,7 @@ public class UndertowOutputStream extends OutputStream implements BufferWritable
                                 break;
                             }
                         }
-                        Channels.writeBlocking(channel, buffers, 0, bufferCount);
+                        writeBlocking(channel, buffers, 0, bufferCount);
                     }
                     buffer.clear();
                 } finally {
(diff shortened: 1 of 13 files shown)


carterkozak:
The proxy test cases are failing on the new tests, I'll investigate when I have a moment.

core/src/main/java/io/undertow/io/UndertowOutputStream.java
```diff
@@ -366,4 +380,12 @@ private ByteBuffer buffer() {
         return this.buffer;
     }
 
+    private long writeBlocking(StreamSinkChannel channel, ByteBuffer[] buffers, int offs, int len) throws IOException {
+        if (writeTimeoutMillis > 0) {
+            return BlockingChannels.writeBlockingOrThrow(
```
carterkozak: bug: this should internally respect the write timeout, but loop as long as bytes are successfully written. Existing callers ignore the result of this method assuming all bytes are fully written.

carterkozak:
I've run benchmarks locally with and without blocking timeouts, differences appear to be within the margin of error so I've not updated existing benchmarks to test this case:
```
Benchmark                                    (blockingTimeout)  (listenerType)  Mode  Cnt     Score     Error  Units
SimpleBenchmarks.benchmarkBlockingEmptyGet            DISABLED            HTTP  avgt    3    30.998 ±   0.678  us/op
SimpleBenchmarks.benchmarkBlockingEmptyGet            DISABLED           HTTPS  avgt    3    61.496 ±   4.240  us/op
SimpleBenchmarks.benchmarkBlockingEmptyGet             ENABLED            HTTP  avgt    3    31.622 ±   1.179  us/op
SimpleBenchmarks.benchmarkBlockingEmptyGet             ENABLED           HTTPS  avgt    3    59.764 ±   2.326  us/op
SimpleBenchmarks.benchmarkBlockingEmptyPost           DISABLED            HTTP  avgt    3    31.605 ±   0.124  us/op
SimpleBenchmarks.benchmarkBlockingEmptyPost           DISABLED           HTTPS  avgt    3    61.610 ±   2.446  us/op
SimpleBenchmarks.benchmarkBlockingEmptyPost            ENABLED            HTTP  avgt    3    31.811 ±   0.538  us/op
SimpleBenchmarks.benchmarkBlockingEmptyPost            ENABLED           HTTPS  avgt    3    59.757 ±  10.410  us/op
SimpleBenchmarks.benchmarkBlockingLargeGet            DISABLED            HTTP  avgt    3  2912.159 ± 418.062  us/op
SimpleBenchmarks.benchmarkBlockingLargeGet            DISABLED           HTTPS  avgt    3  4925.408 ± 741.017  us/op
SimpleBenchmarks.benchmarkBlockingLargeGet             ENABLED            HTTP  avgt    3  2873.870 ± 416.579  us/op
SimpleBenchmarks.benchmarkBlockingLargeGet             ENABLED           HTTPS  avgt    3  4924.962 ± 482.183  us/op
SimpleBenchmarks.benchmarkBlockingLargePost           DISABLED            HTTP  avgt    3   612.207 ±  35.712  us/op
SimpleBenchmarks.benchmarkBlockingLargePost           DISABLED           HTTPS  avgt    3  2629.098 ± 145.288  us/op
SimpleBenchmarks.benchmarkBlockingLargePost            ENABLED            HTTP  avgt    3   615.682 ±  58.484  us/op
SimpleBenchmarks.benchmarkBlockingLargePost            ENABLED           HTTPS  avgt    3  2614.834 ± 130.408  us/op
```

carterkozak:
@fl4via I think I have a simpler solution using a custom HttpHandler to apply request and response conduit wrappers. This way we can measure time spent in awaitReadable/awaitWritable, and reset the timer when a read/write method returns non-zero.
This approach does not require any new UndertowOptions, nor modification of existing code. Unlike the async-aware implementations, this approach doesn't risk violating the undertow thread model because callers of awaitWritable must own the conduit.
The drawback is that we end up calling `System.nanoTime()` twice for each await* invocation, which is about double what we expect using this approach, but I think the simplicity is worth the trade-off. I can put together a PR with this proposal for comparison.

carterkozak:
PR is up: https://github.com/undertow-io/undertow/pull/837

fl4via:
@carterkozak I'll be tagging 2.1.0.CR1 this week, soon I hope. Both PRs will be reviewed and I'll give you feedback as soon as I finish working on some new features for Undertow/WF

carterkozak:
Thanks :-)

fl4via:
@carterkozak  I am approving the general idea proposed by this commit. However, I'll have to follow a process since it adds a new feature to Undertow. You can expect it to be released within a month, and probably we will tag a CR before that.

carterkozak:
Thanks, I’m also happy to close this out in favor of the approach used in #837 which I expect to be easier to support, and matches the undertow handler wrapper architecture more cleanly.

PR #831, opened 2019-12-11
UNDERTOW-1627 gzip encoding is not enabled by RequestEncodingHandler$…

msfm:
…Builder

Add gzip encoding in RequestEncodingHandler$Builder as it enables only
the deflate encoding. Also, rename the handler builder name to a more
appropriate one and add a missing RequestEncodingHandler$Builder
definition in a provider configuration file.

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/encoding/RequestEncodingHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/encoding/RequestEncodingHandler.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.xnio.conduits.StreamSourceConduit;
+import io.undertow.conduits.GzipStreamSourceConduit;
 import io.undertow.conduits.InflatingStreamSourceConduit;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/server/handlers/encoding/RequestEncodingHandler.java
```diff
@@ -84,7 +85,7 @@ public HttpHandler getNext() {
 
         @Override
         public String name() {
-            return "uncompress";
+            return "uncompress-request";
```
fl4via: I'm rejecting this change because this name is part of Undertow documentation here: http://undertow.io/undertow-docs/undertow-docs-2.0.0/predicates-attributes-handlers.html
While we can assume that nobody was using this handler because it was missing from the HandlerBuilder file, I think we must stick to what the documentation says, at least for version 2.0.x

fl4via: I'm updating this PR since I would like to merge it before tagging the next release. Thanks for the fix btw!

PR #832, opened 2019-12-13
UNDERTOW-1464 Proper handling of matrix parameters

chrsoo:
Hi @stuartwdouglas, @fl4via,

Please review and provide feedback on my first stab at resolving the [UNDERTOW-1464](https://issues.redhat.com/browse/UNDERTOW-1464)! Some notes on the code
* The PR puts all matrix parameters in the path parameter collection and makes no difference between the two `p1` matrix parameters in the following example
  ```
  /some;p1=v1/path;p1=v2
  ```
  An alternative implementation could take into account the fact that these are really two separate parameters, perhaps by using the canonical path as a namespace which would give the following two variable names:
  ```
  /some;p1
  /some/path;p1
  ```
* The current PR does not include support for comma separated values, I will add that shortly if you think this PR is going anywhere.
  ```
  /some;p1=v1,v2
  ```
* The `result.getRequestPath()` returns the canonical path without the matrix parameters and the `result.getRequestURI()` returns the URI including the matrix parameters. Maybe the path should include the parameters as well? Not sure what's best, thoughts?

References:
* [Matrix URLs](https://www.w3.org/DesignIssues/MatrixURIs.html) as first discussed by Tim Berners Lee...

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/http/HttpRequestParser.java
+++ b/core/src/main/java/io/undertow/server/protocol/http/HttpRequestParser.java
@@ -687,9 +687,16 @@ public abstract class HttpRequestParser {
                     state.segmentStart = stringBuilder.length();
                     state.parseState = ParseState.PATH;
                     return;
+                } else if (next == ',') {
+                    if(param == null) {
+                        throw UndertowMessages.MESSAGES.failedToParsePath();
+                    } else {
+                        // parse value and add param
+                        exchange.addPathParam(param, decode(stringBuilder.substring(pos), urlDecodeRequired, state, true, true));
+                        pos = stringBuilder.length() + 1;
+                    }
                 }
                 stringBuilder.append(next);
-
             }
 
         }
--- a/core/src/main/java/io/undertow/server/protocol/http/HttpRequestParser.java
+++ b/core/src/main/java/io/undertow/server/protocol/http/HttpRequestParser.java
@@ -640,9 +640,10 @@ public abstract class HttpRequestParser {
                 }
                 finalizePath(state, exchange, urlDecodeRequired);
                 if(next == '?') {
-                    state.parseState = ParseState.QUERY_PARAMETERS;
+                    state.state = ParseState.QUERY_PARAMETERS;
+                    handleQueryParameters(buffer, state, exchange);
                 } else {
-                    state.parseState = ParseState.VERSION;
+                    state.state = ParseState.VERSION;
                 }
                 state.stringBuilder.setLength(0);
                 state.pos = 0;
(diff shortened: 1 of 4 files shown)


undertow-pull-request:

fl4via:
This is ok to test

chrsoo:
@fl4via it looks like the Team City build failed but I can't see any details as I do not have access. Please tell me what to fix, if anything, in order for this PR to make it into production!

ctomc:
@chrsoo on login screen, just click "login as guest"

spyrkob:
I tried to run tests locally - most of the failing tests from CI seem to be passing, but there were a few issues. I think https://github.com/undertow-io/undertow/commit/ba300ae26234a899fdfac400228a7f5105c5e721 should fix those. 

@fl4via in https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/server/session/PathParameterSessionConfig.java#L164 the session parameter encoding seems to be using '&' to append additional path parameters - is that a correct behaviour or should it be using ';' instead?

chrsoo:
> @fl4via in https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/server/session/PathParameterSessionConfig.java#L164 the session parameter encoding seems to be using '&' to append additional path parameters - is that a correct behaviour or should it be using ';' instead?

AFAIK there is no formal spec of *matrix* parameters but in the examples of the [Matrix URIs](https://www.w3.org/DesignIssues/MatrixURIs.html) document where Tim Berners-Lee introduced the concept, he uses semi colons as separators.

This said I assume there might be considerations for backwards compatibility and maybe we need to support both the ampersand and semi colon as a separator?

Regarding the difference between path and matrix parameters cf. my [stack overflow response](https://stackoverflow.com/a/59254251/51016) that I posted some time ago!

@spyrkob - in any case, thanks for looking into this patch and helping it through the process!

fl4via:
@chrsoo  good point about compatibility. I'm reviewing all this PR and I might add a few changes as well. I'll update the branch accordingly. Hopefully we will see this one merged pretty soon.

fl4via:
> @chrsoo good point about compatibility. I'm reviewing all this PR and I might add a few changes as well. I'll update the branch accordingly. Hopefully we will see this one merged pretty soon.

I'll just ignore that because the & is in the generated output, so I'm going with what is right which is the ';'

fl4via:
I'm finding it odd that two new failures in Linux that were basically random before are not persistently occuring, I'm double checking this before merging

chrsoo:
@fl4via @spyrkob - thanks for fixing and seeing this PR through!

PR #837, opened 2020-01-03
UNDERTOW-1622: Implement blocking IO timeouts

carterkozak:
Implemented a new BlockingTimeoutHandler which can be configured
with I/O timeouts for blocking read and write operations.

This approach limits the time that may be spent in awaitReadable
or awaitWritable to apply to only blocking operations without requiring
invasive code change.

### DIFF
--- a/core/src/test/java/io/undertow/server/handlers/BlockingWriteTimeoutHandlerTestCase.java
+++ b/core/src/test/java/io/undertow/server/handlers/BlockingWriteTimeoutHandlerTestCase.java
@@ -1,0 +1,105 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package io.undertow.server.handlers;
+
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.testutils.DefaultServer;
+import io.undertow.testutils.HttpOneOnly;
+import io.undertow.testutils.ProxyIgnore;
+import io.undertow.testutils.TestHttpClient;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/undertow/server/handlers/BlockingTimeoutHandler.java
```diff
@@ -0,0 +1,367 @@
...
+/**
+ * BlockingTimeoutHandler allows configurable blocking I/O timeouts for read
+ * and write operations within an exchange.
+ * Unlike Options.READ_TIMEOUT and Options.WRITE_TIMEOUT this only applies to
+ * blocking operations which can be helpful to prevent the worker pool from
+ * becoming saturated when clients stop responding.
+ *
+ * @author Carter Kozak
+ */
+public final class BlockingTimeoutHandler implements HttpHandler {
```
carterkozak: Open question: should this be split into separate BlockingReadTimeoutHandler and BlockingWriteTimeoutHandler?

carterkozak:
Failing jdk11 tests appear to be unrelated to this change, but may be related to:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8198562
similar netty issue:
https://github.com/netty/netty/issues/8697#issuecomment-450619821

core/src/main/java/io/undertow/server/handlers/BlockingTimeoutHandler.java
carterkozak: I think this makes more sense to split into separate handlers for read and write timeouts, that way each one is responsible for exactly one thing.

core/src/main/java/io/undertow/server/handlers/BlockingTimeoutHandler.java
```diff
@@ -0,0 +1,422 @@
...
+        public void awaitReadable(long duration, TimeUnit unit) throws IOException {
+            long startTime = System.nanoTime();
+            long requestedNanos = unit.toNanos(duration);
+            try {
+                delegate.awaitReadable(Math.min(requestedNanos, remaining), TimeUnit.NANOSECONDS);
+            } finally {
+                remaining -= (System.nanoTime() - startTime);
+            }
+            if (remaining < 0) {
+                IoUtils.safeClose(serverConnection);
```
carterkozak: Might be helpful to log something at debug level here and in the write timeout conduit when the timeout is reached.

core/src/test/java/io/undertow/server/handlers/BlockingWriteTimeoutHandlerTestCase.java
```diff
@@ -0,0 +1,105 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
fl4via, line 3: I'll update these headers before merging.

fl4via (changes requested):

fl4via (approved):

PR #838, opened 2020-01-03
[UNDERTOW-1623] Fix the deadlock by invoking lastDataRead at IO thread

fl4via:
Jira: https://issues.redhat.com/browse/UNDERTOW-1623

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
@@ -388,7 +388,8 @@ public abstract class AbstractFramedChannel<C extends AbstractFramedChannel<C, R
             } else if (read == -1 && !hasData) {
                 forceFree = true;
                 readChannelDone = true;
-                lastDataRead();
+                getIoThread().execute(()->{
+                    lastDataRead();});
                 return null;
             } else if(isLastFrameReceived() && frameDataRemaining == 0) {
                 //we got data, although we should have received the last frame


fl4via:
This is an alternative fix for #828

fl4via:
This is ok to test

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
```diff
@@ -388,7 +388,7 @@ public synchronized R receive() throws IOException {
             } else if (read == -1 && !hasData) {
                 forceFree = true;
                 readChannelDone = true;
-                lastDataRead();
+                getIoThread().execute(()->{lastDataRead();});
```
carterkozak: We can slightly simplify to use a method reference
```suggestion
                getIoThread().execute(this::lastDataRead);
```

Should Http2Channel declare `lastDataRead` volatile?

fl4via: @carterkozak good catch! I'll update the PR

fl4via: @carterkozak as a second thought, I'm closing this PR and sticking with the original one. I think that asynchronously invoking lastDataRead from receive might have collateral effects on receive logic. The reason is that receive uses isLastFrameReceived() to decide if we are at last frame and  Http2Channel.isLastFrameReceived returns lastDataRead. So, by doing it asynchronously we risk a receive missing a true isLastFrameReceived.

fl4via:
I'm closing this PR and sticking with #828 . I think that asynchronously invoking lastDataRead from receive might have collateral effects on receive logic. The reason is that receive uses isLastFrameReceived() to decide if we are at last frame and Http2Channel.isLastFrameReceived returns lastDataRead. So, by doing it asynchronously we risk a receive missing a true isLastFrameReceived

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
carterkozak: That makes sense, I agree that the synchronized approach is safer. Thank you for fixing the bug!

PR #840, opened 2020-01-15
[UNDERTOW-1637] Redirect Form auth on root context without trailing s…

spyrkob:
…lash

Issue: https://issues.redhat.com/browse/UNDERTOW-1637

@fl4via could you please review? Alternatively the redirect could be handled in https://github.com/undertow-io/undertow/blob/master/servlet/src/main/java/io/undertow/servlet/handlers/ServletInitialHandler.java#L140, but doing that would undo https://issues.redhat.com/browse/UNDERTOW-1578 for requests to root context

### DIFF
--- a/servlet/src/test/java/io/undertow/servlet/test/security/form/FormAuthenticationRootContextRedirectTestCase.java
+++ b/servlet/src/test/java/io/undertow/servlet/test/security/form/FormAuthenticationRootContextRedirectTestCase.java
@@ -1,0 +1,124 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


spyrkob:
retest this please

servlet/src/test/java/io/undertow/servlet/test/security/form/FormAuthenticationRootContextRedirectTestCase.java
```diff
@@ -0,0 +1,123 @@
+/**
```
fl4via: I think we are missing the first block in the license header? Is there a reason for this or was it mistakenly removed?

spyrkob: Looks like I had wrong template in the IDE, I'll change that

spyrkob:
@fl4via the license is updated now. Apart from that is  the PR looking OK?

PR #842, opened 2020-01-20
UNDERTOW-1639: AbstractFramedStreamSourceChannel notifies on close

carterkozak:
Previously when a framed channel was closed from an I/O thread while
a worked waited for blocking operations, the blocking thread could
become stuck forever.

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
@@ -83,6 +83,7 @@ public abstract class AbstractFramedStreamSourceChannel<C extends AbstractFramed
     private long frameDataRemaining;
 
     private final Object lock = new Object();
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
```diff
@@ -646,6 +646,10 @@ public synchronized void close() {
                     closeListeners[i].handleEvent(this);
                 }
             }
+            // UNDERTOW-1639: Close may be called from an I/O thread while a worker is blocked on awaitReadable.
+            // Once the channel is closed, callers must be awoken. Note that this does not check 'waiters' because
+            // the value is not volatile and could result in a race.
+            lock.notifyAll();
```
carterkozak: This could live in `channelForciblyClosed`, but I'm not sure there's much downside in calling it for each invocation. This way should be safer.

PR #845, opened 2020-01-21
Tests cycle through ports to avoid 'BindException: Address already in use'

carterkozak:
Note: I don't have a great idea how internal infrastructure works, it's possible this will cause unforeseen issues.

Goal is to reduce test flakes when the server port isn't reset quickly enough for the next test, however it's possible this could mask problems if listeners are not shut down. This change would not fix flakes on test infrastructure that explicitly sets `default.server.port`.


carterkozak:
I think there's a legitimate bug here, I have an undertow based project with similar failures.

PR #847, opened 2020-01-22
UNDERTOW-1643: Prevent deadlock in AbstractFramedStreamSinkChannel

carterkozak:
Close uses double-check locking with the encapsulated lock object
instead of relying on the object level lock which risks deadlocks.

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
@@ -527,12 +527,16 @@ public abstract class AbstractFramedStreamSinkChannel<C extends AbstractFramedCh
     }
 
     @Override
-    public synchronized void close() throws IOException {
+    public void close() throws IOException {
         if(fullyFlushed || anyAreSet(state, STATE_CLOSED)) {
             return;
         }
         try {
             synchronized (lock) {
+                // Double check to avoid executing the the rest of this method multiple times
+                if(fullyFlushed || anyAreSet(state, STATE_CLOSED)) {
+                    return;
+                }
                 state |= STATE_CLOSED;
             }
             if(writeBuffer != null) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
```diff
@@ -527,12 +527,16 @@ public boolean isOpen() {
     }
 
     @Override
-    public synchronized void close() throws IOException {
```
carterkozak: @fl4via I didn't see anywhere else that synchronizes on the instance so I assumed that this was added to prevent multiple invocations of the close method body, but I could be mistaken.

fl4via: @carterkozak it was make synchronized to protect from the removal of synchronization from AbstractFramedChannel.receive()
That said, maybe this fix also works. Let's see first what is made of https://issues.redhat.com/browse/WFLY-12822, to see if the previous fix of UNDERTOW-1643 will be kept

carterkozak: Thanks for the information!

If #852 merges, I think we'll still want to keep the changes to check STATE_CLOSED under the lock in order to avoid multiple concurrent executions.

Tagging @ropalka for context as well

ropalka: We must avoid as much locks as possible. Introducing synchronization on AbstractFramedStreamSinkChannel object itself was not ideal step forward to fix deadlock IMHO.

carterkozak: Yes, I completely agree, I'm referring to the existing locking in the close method on line 534. On master two threads can call close at the same time, both see that the channel is not closed, and each lock before setting `state |= STATE_CLOSED;`. Without either a check under synchronization or a compare-and-swap check, it will still be possible to run the body of the close method twice, which we expect to be run once.

fl4via: @carterkozak I agree with you on that. We need to fix that as well.

fl4via:
Jira: https://issues.redhat.com/browse/UNDERTOW-1643

PR #849, opened 2020-01-27
[UNDERTOW-1640] Sequenced http/2 fixes to avoid merge conflicts

carterkozak:
[UNDERTOW-1640](https://issues.redhat.com/browse/UNDERTOW-1640)

```
    UNDERTOW-1640: Fix AbstractFramedStreamSourceChannel buffer access
    
    When streams are canceled unexpectedly (http/2 rst_stream while blocking
    operations execute, for example) interaction between close being called
    on an I/O thread and blocking work on a task thread was not properly
    handled.
    
    This manifested in several ways:
    1. NPE in AbstractFramedStreamSourceChannel.read
    2. "Buffer has already been freed" error in read
    3. "BufferOverflowException" from read operations
    4. modification of a returned buffer that is now
       used by another operation (data corruption or
       protocol error)
```

```
    UNDERTOW-1640: Fix AbstractFramedStreamSinkChannel buffer access
    
    When streams are canceled unexpectedly (http/2 rst_stream while blocking
    operations execute, for example) interaction between close being called
    on an I/O thread and blocking work on a task thread was not properly
    handled.
    
    Unlike the AbstractFramedStreamSourceChannel case I have not
    forced a reproducer for invalid buffer access, it's a bit tricker
    to hit without spending a lot of iterations, but the result would
    be the same.

```

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
@@ -365,17 +365,17 @@ public abstract class AbstractFramedStreamSinkChannel<C extends AbstractFramedCh
                 state |= STATE_CLOSED;
                 return true;
             }
-        }
-        if (anyAreSet(state, STATE_WRITES_SHUTDOWN) && !finalFrameQueued) {
-            queueFinalFrame();
-            return false;
-        }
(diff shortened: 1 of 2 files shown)


fl4via (approved):

fl4via (approved):

fl4via:
Unfortunately we are seeing consistent failures, I'll have to leave this commit out of next release and investigate further the failures.

PR #851, opened 2020-01-28
[UNDERTOW-1623] Make AbstractFramedChannel.queueFrame delegate safeCl…

fl4via:
…ose from IoThread

PR #852, opened 2020-01-28
[UNDERTOW-1623] Undertow Deadlock

ropalka:
https://issues.redhat.com/browse/UNDERTOW-1623

In order to avoid deadlocks we cannot call alien or potentially dangerous methods
while lock is being held. We have to hold as few locks as possible.

This PR supersedes:
 * https://github.com/undertow-io/undertow/pull/851
 * https://github.com/undertow-io/undertow/pull/847

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
@@ -262,9 +262,9 @@ public abstract class AbstractFramedStreamSinkChannel<C extends AbstractFramedCh
                 state |= STATE_FIRST_DATA_WRITTEN;
                 state |= STATE_WRITES_SHUTDOWN; // Mark writes as shutdown as well, since we want that set prior to queueing
                 finalFrameQueued = true;
-                channel.queueFrame((S) this);
-            }
+            } else return;
         }
+        channel.queueFrame((S) this);
     }
 
     protected boolean isFinalFrameQueued() {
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
```diff
@@ -240,12 +240,12 @@ public void run() {
 
     @Override
     public void shutdownWrites() throws IOException {
+        // Queue prior to shutting down writes, since we might send the write buffer
```
carterkozak: Do you think we should add an assert here to validate `lock` is not already held?

ropalka: It is not necessary @carterkozak - lock is private variable in AbstractFramedStreamSinkChannel and so external classes using this method cannot obtain the lock.

carterkozak: I was thinking it could prevent regressions if this class is updated in the future, but I defer to your judgement.

ropalka: In such context (defensive programming) it would make sense of course.

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
```diff
@@ -240,12 +240,12 @@ public void run() {
...
+        queueFinalFrame();
```
fl4via: This is not the same as before. Before, inside the lock, we would skip queueFinalFrame if (anyAreSet(state, STATE_WRITES_SHUTDOWN) || broken). Now, we are always invoking queueFinalFrame no matter what.

ropalka: That's not the problem @fl4via . I analyzed the code and I can guarantee that behaviour is persisted. This is because 
'broken' flag is also checked inside queueFinalFrame() method. If it is set to true, method will not execute. What about anyAreSet(state, STATE_WRITES_SHUTDOWN) check
that also has no effect on queueFinalFrame() method. It is because STATE_WRITES_SHUTDOWN is 'written' only in shutdownWrites() and queueFinalFrame() methods.

fl4via: @ropalka I see what you mean now, I'm approving this PR.

fl4via:
@ropalka for readability, I'm not condensing the PR into a single commit. Can you please add the Jira number to the other two commit messages before I merge?

ropalka:
Commits fixed as requested @fl4via

PR #853, opened 2020-01-28
[UNDERTOW-1649] Fix AbstractFramedStreamSourceChannel.awaitReadable() guard methods

ropalka:
https://issues.redhat.com/browse/UNDERTOW-1649

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
@@ -342,43 +342,47 @@ public abstract class AbstractFramedStreamSourceChannel<C extends AbstractFramed
 
     @Override
     public void awaitReadable() throws IOException {
-        if(Thread.currentThread() == getIoThread()) {
+        if (Thread.currentThread() == getIoThread()) {
             throw UndertowMessages.MESSAGES.awaitCalledFromIoThread();
         }
-        if (data == null && pendingFrameData.isEmpty()) {
-            synchronized (lock) {
-                if (data == null && pendingFrameData.isEmpty()) {
-                    try {
-                        waiters++;
-                        lock.wait();
-                    } catch (InterruptedException e) {
-                        Thread.currentThread().interrupt();
-                        throw new InterruptedIOException();
-                    } finally {
-                        waiters--;
-                    }
+        synchronized (lock) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
```diff
@@ -342,43 +342,43 @@ protected boolean isLastFrame() {
...
-                        waiters++;
-                        lock.wait(timeUnit.toMillis(l));
-                    } catch (InterruptedException e) {
-                        Thread.currentThread().interrupt();
-                        throw new InterruptedIOException();
-                    } finally {
-                        waiters--;
-                    }
+        synchronized (lock) {
+            while (data == null && pendingFrameData.isEmpty()) {
```
carterkozak: I don't think this is correct, this may result in waiting longer than the intended duration

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
```diff
@@ -342,43 +342,43 @@ protected boolean isLastFrame() {
...
-                        waiters++;
-                        lock.wait();
-                    } catch (InterruptedException e) {
-                        Thread.currentThread().interrupt();
-                        throw new InterruptedIOException();
-                    } finally {
-                        waiters--;
-                    }
+        synchronized (lock) {
+            while (data == null && pendingFrameData.isEmpty()) {
```
carterkozak: I'm not sure this is strictly necessary given awaitReadable and awaitWritable are allowed to return spuriously.

This change will also conflict with #842

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
ropalka: Thanks for spotting that out - fixed.

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
ropalka: You are right @carterkozak - these methods can return spuriously. 
I had to check API definition first. I can close this PR evidently :)

PR #857, opened 2020-02-26
UNDERTOW-1661 Fix Exchange already complete while rendering a JSP

foch:
We now test in MetricsHandler that all conditions are met before adding an ExchangeCompletionListener to avoid an unwanted failure.

### DIFF
--- a/core/src/main/java/io/undertow/server/HttpServerExchange.java
+++ b/core/src/main/java/io/undertow/server/HttpServerExchange.java
@@ -937,7 +937,7 @@ public final class HttpServerExchange extends AbstractAttachable {
 
 
     public HttpServerExchange addExchangeCompleteListener(final ExchangeCompletionListener listener) {
-        if(isComplete() || this.exchangeCompletionListenersCount == -1) {
+        if(isComplete() || hasListenersBeenInvoked()) {
             throw UndertowMessages.MESSAGES.exchangeAlreadyComplete();
         }
         final int exchangeCompletionListenersCount = this.exchangeCompletionListenersCount++;
@@ -951,6 +951,10 @@ public final class HttpServerExchange extends AbstractAttachable {
         }
         exchangeCompleteListeners[exchangeCompletionListenersCount] = listener;
         return this;
+    }
+
+    public boolean hasListenersBeenInvoked() {
+        return this.exchangeCompletionListenersCount == -1;
     }
 
     public HttpServerExchange addDefaultResponseListener(final DefaultResponseListener listener) {
--- a/core/src/main/java/io/undertow/server/handlers/MetricsHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/MetricsHandler.java
@@ -50,7 +50,7 @@ public class MetricsHandler implements HttpHandler {
 
     @Override
     public void handleRequest(HttpServerExchange exchange) throws Exception {
-        if(!exchange.isComplete()) {
+        if(!exchange.isComplete() && !exchange.hasListenersBeenInvoked()) {
             final long start = System.currentTimeMillis();
             exchange.addExchangeCompleteListener(new ExchangeCompletionListener() {
                 @Override


undertow-pull-request:

iweiss:
this is ok to test

bmaxwell:
@fl4via when you get a chance can you check this? There is a patch request for it.

core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -953,6 +953,10 @@ public HttpServerExchange addExchangeCompleteListener(final ExchangeCompletionLi
         return this;
     }
 
+    public boolean isExchangeCompletionListenersEmpty() {
+        return this.exchangeCompletionListenersCount == -1;
```
carterkozak: I don't think this returns true when there are no listeners (empty), only once listeners have been executed. I'm curious if you have a minimal reproducer (something like a unit test) to produce this behavior, it sounds like exchange.isComplete should return true in this case rather than requiring two separate checks.

foch: Hello @carterkozak 
thanks for the heads up on what the check is actually doing. I have renamed the method accordingly.
 I'm afraid I don't have a minimal reproducer, as the error happens randomly on our customer's webapp, when rendering a JSP and refreshing the page several times. It's probably a concurrent error.
 Fell free to propose a new name for the method. Alternatively, we could patch isComplete() to add the check inside, but I don't have the knowledge on this project to foresee the impacts.

fl4via: @carterkozak  thanks for looking into this.
You're right, it is not checking when there are no listeners. The problem here is that -1 is marked as soon as listeners are going to be invoked, we need to make sure -1 is marked only when all listeners have been invoked instead.

fl4via: @foch  were you able to test this patch with the customer's webapp to validate the fix? I'm not certain this patch solves the issue.

foch: Yes I managed to patch directly the Undertow jar file in our JBoss EAP docker image, and it fixes the issue.

fl4via: @foch thanks for the info!

carterkozak: Would that allow listeners to be added from other listeners? That could become difficult to reason about, and would require the mechanism to be reimplemented.
Adding a new API on HttpServerExchange could be helpful more broadly: 

```java
/** Returns true if the listener was successfully added, false if this exchange has already completed. */
public boolean tryAddExchangeCompleteListener(final ExchangeCompletionListener listener)
```

I don't know the process for adding new APIs to core components, if that's problematic the MetricsHandler could be updated to try/catch in the interim.

fl4via: @foch please, update the method name to hasListenersBeenInvoked, since right now we have no guarantee that they all have completed execution, just that they have been invoked

fl4via:
This is ok to test

core/src/main/java/io/undertow/server/HttpServerExchange.java
foch: Done :)

carterkozak: I'm sorry to jump into this discussion as an outsider, you've reminded me of a few bugs that I've had to unwind (and likely also caused) in a project I'm responsible for! Do you think something along these lines would be helpful as an alternative? #864

fl4via: @carterkozak  sorry for not answering your previous comment, somehow I missed it.
We cannot add new methods to a core component on a minor release such as 2.0.30.Final
Still, the suggestion is going to be taken into account for a future release.

core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -953,6 +953,10 @@ public HttpServerExchange addExchangeCompleteListener(final ExchangeCompletionLi
         return this;
     }
 
+    public boolean hasListenersBeenInvoked() {
```
fl4via: Oh, just now I noticed this. :-( This method is public. I won't merge this PR for now, this will be reviewed further before we get to a decision.

fl4via:
We are fixing this with PR #867

PR #858, opened 2020-02-28
UNDERTOW-1664: Prevent errors when http/2 channels are terminated

carterkozak:
This can be reproduced using
https://github.com/carterkozak/http2-reproducer-undertow
specifically the second test case:
https://github.com/carterkozak/http2-reproducer-undertow/tree/master/src/main/java/net/ckozak/repro/two

This PR provides two changes:

1. Http/2 channels may be closed prior to the http2 server connection
    object. In this case an error is no longer logged, and the exchange
    is completed as though the connection is closed. This fixes all instances
    of the bug that I'm aware of. I can no longer reproduce the bug with this fix.

2. Exchange completion listeners are invoked in the unexpected case
    to avoid resource leaks, and allow request logging to execute
    as expected. This is defensive programming in case another issue of this
    type is introduced, this way we log an error, but allow resources to be returned.

The previous behavior resulted in result leaks and lack of request logging because exchange completion handlers were not logged in this case.

### DIFF
--- a/core/src/main/java/io/undertow/server/HttpServerExchange.java
+++ b/core/src/main/java/io/undertow/server/HttpServerExchange.java
@@ -1724,6 +1724,13 @@ public final class HttpServerExchange extends AbstractAttachable {
                     getResponseHeaders().put(Headers.CONTENT_LENGTH, "0");
                 }
                 getResponseChannel();
+            } else if (anyAreClear(state, FLAG_RESPONSE_TERMINATED) && !responseChannel.isOpen()) {
+                // UNDERTOW-1664: Http/2 response channels may be closed prior to the connection. There's
+                // no reason to attempt to flush a response for a closed channel but we must ensure
+                // the listeners have been invoked.
+                invokeExchangeCompleteListeners();
+                IoUtils.safeClose(connection);
+                return;
             }
             responseChannel.shutdownWrites();
             if (!responseChannel.flush()) {


core/src/main/java/io/undertow/UndertowLogger.java
```diff
@@ -393,7 +393,7 @@ void nodeConfigCreated(URI connectionURI, String balancer, String domain, String
     @Message(id = 5084, value = "Attempted to write %s bytes however content-length has been set to %s")
     IOException dataLargerThanContentLength(long totalToWrite, long responseContentLength);
 
-    @LogMessage(level = ERROR)
+    @LogMessage(level = WARN)
```
carterkozak: Thoughts on moving this down to INFO level given we know it occurs when a rst_stream frame is received?

carterkozak:
The test failures appear to have started when #832 merged

core/src/main/java/io/undertow/UndertowLogger.java
carterkozak: No longer necessary, I root caused and fixed the bug.

core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -1716,6 +1716,13 @@ private void closeAndFlushResponse() {
                     getResponseHeaders().put(Headers.CONTENT_LENGTH, "0");
                 }
                 getResponseChannel();
+            } else if (anyAreClear(state, FLAG_RESPONSE_TERMINATED) && !responseChannel.isOpen()) {
+                // UNDERTOW-1664: Http/2 response channels may be closed prior to the connection. There's
+                // no reason to attempt to flush a response for a closed channel but we must ensure
+                // the listeners have been invoked.
+                invokeExchangeCompleteListeners();
+                IoUtils.safeClose(connection);
+                return;
```
carterkozak: @fl4via I'd appreciate feedback on this PR when you have time. This block doesn't feel like the best solution, but I was not able to find the call site which closes the response channel leaving an exchange in an incomplete state. If this block is removed from the PR the server logs a lot of errors complaining that the connection was closed uncleanly, so I opted to include handling for the known case so the logging is available for remaining unexpected cases.
Regardless of this block, I think it's critical to invoke completion listeners, otherwise request logging will not occur and resources may not be released (RequestBufferingHandler request buffers, for example).

fl4via:
@carterkozak  Can you please let me know which test failure you think started after that PR? If you noticed a specific failure, I'll review it right away, I didn't notice a new failure at that time when merging #832

carterkozak:
Both `ServletSessionTestCase.testSessionConfigNoCookies` and `ServletURLRewritingSessionTestCase.testURLRewriting` with the combination of `-Dtest.proxy=true` and either `-Dtest.ajp=true` or `-Dtest.h2=true`

carterkozak:
To clarify, those test failures are causing this PR to fail CI, but aren't related to this PR.

This PR is based on pager alerts from a service that logged too many errors over a ten minute window due to the log line with ID 5085 `Connection %s for exchange %s was not closed cleanly, forcibly closing connection`

core/src/main/java/io/undertow/server/HttpServerExchange.java
ropalka: Hi @carterkozak - I believe this https://github.com/undertow-io/undertow/pull/867 fixed this problem already.
Could you please test the problematic scenario you are facing with latest Undertow 2.0.30.Final release 
and confirm the problem was solved?
Just FYI ^^^ @fl4via

carterkozak: Thanks for the ping! I thought this could be related as well, but the issue is still reproducible with this commit (I've re-validated using 2.0.30.Final as well).

You can try it yourself using this reproducer with the `Client` and `Server` classes in the `two` package :-) It reproduces with `one` as well, but that's more likely to hit other bugs, and I don't want to conflate the two.

https://github.com/carterkozak/http2-reproducer-undertow

carterkozak: Note that in my scenario the Http2ServerConnection is still considered open despite the stream being closed.

fl4via (approved):

PR #859, opened 2020-03-04
UNDERTOW-1667 Change AjpRequestParser to ignore unknown request attri…

msfm:
…butes and add a new UndertowOption AJP_ALLOWED_REQUEST_ATTRIBUTES_PATTERN

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/ajp/AjpOpenListener.java
+++ b/core/src/main/java/io/undertow/server/protocol/ajp/AjpOpenListener.java
@@ -55,6 +55,8 @@ import static io.undertow.UndertowOptions.URL_CHARSET;
  */
 public class AjpOpenListener implements OpenListener {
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/undertow/server/protocol/ajp/AjpOpenListener.java
```diff
@@ -55,6 +55,8 @@
  */
 public class AjpOpenListener implements OpenListener {
 
+    private static final String DEFAULT_AJP_ALLOWED_REQUEST_ATTRIBUTES_PATTERN = System.getProperty("io.undertow.ajp.allowedRequestAttributesPattern");
```
jamezp: Should this use a privileged action?

fl4via: @jamezp in fact, I have missed that fact when reviewing this earlier!
We do have a SecurityActions class for this, I'll update the PR since @msfm is in another time zone

msfm: Thanks for pointing this out. I've just found that privileged action is not used when executing System.getPropery() in other places. So, I think we will also need to update others later.

PR #862, opened 2020-03-08
Http session to request/response references

evernat:
Since 0c0abf2, http sessions keep reference to request and response via servletRequestContext.
Because an http session can be kept as a reference by the webapp for some time or for a long time after the end of the request, it would be better for memory that an http session does not keep references to request and response.

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/spec/HttpSessionImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/HttpSessionImpl.java
@@ -18,6 +18,7 @@
 
 package io.undertow.servlet.spec;
 
+import java.lang.ref.WeakReference;
 import java.security.PrivilegedAction;
 import java.util.Enumeration;
 import java.util.HashSet;
@@ -49,13 +50,13 @@ public class HttpSessionImpl implements HttpSession {
     private final ServletContext servletContext;
     private final boolean newSession;
     private volatile boolean invalid;
-    private final ServletRequestContext servletRequestContext;
+    private final WeakReference<ServletRequestContext> servletRequestContextRef;
 
     private HttpSessionImpl(final Session session, final ServletContext servletContext, final boolean newSession, ServletRequestContext servletRequestContext) {
         this.session = session;
         this.servletContext = servletContext;
         this.newSession = newSession;
-        this.servletRequestContext = servletRequestContext;
+        this.servletRequestContextRef = new WeakReference<>(servletRequestContext);
     }
 
     public static HttpSessionImpl forSession(final Session session, final ServletContext servletContext, final boolean newSession) {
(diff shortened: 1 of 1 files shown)


undertow-pull-request:

stuartwdouglas:
Weak references are almost never a good idea. 

Are you actually seeing this as a real problem? The underlying long term term storage is io.undertow.server.session.Session, the HttpSession implementations are not long lived.

evernat:
@stuartwdouglas 
Yes, I am seeing this as a real problem, when using javamelody for example.

>the HttpSession implementations are not long lived.

What makes you think that? Is it an assumption on your part?
HttpSessions often live 30 minutes or sometimes a few hours in my point of view.

> The servlet container uses this interface [HttpSession] to create a session between an HTTP client and an HTTP server. The session persists for a specified time period, across more than one connection or page request from the user. A session usually corresponds to one user, who may visit a site many times. The server can maintain a session in many ways such as using cookies or rewriting URLs. 

according to https://javadoc.io/doc/javax.servlet/javax.servlet-api/latest/javax/servlet/http/HttpSession.html

stuartwdouglas:
The actual data is stored in io.undertow.server.session.Session, the HttpSession is not the persistent data store, but a per request facade.

OBreidenbach:
We found th problem of larger sessions when using java-melody after upgrading wildfly 10.1 (untertow 1.4.0) to wildfly 18 (undertow 2.0.26).
As the https://github.com/undertow-io/undertow/commit/0c0abf223ab2d967cf0a42487b70cb450e8e0bc7 was already included in undertow 1.2 it can not be the only reason for the problem. 
I did browse through the last commits and maybe https://github.com/undertow-io/undertow/commit/439001a0ca1fb4b0f44d63b06a66fe9000899202 causes the different behavior?

OBreidenbach:
We did revert https://github.com/undertow-io/undertow/commit/439001a0ca1fb4b0f44d63b06a66fe9000899202 for undertow release 2.0.27. With this patch undertow the needed memory did shrink a lot. So the increased memory usage together with usage of javamelody is because of this commit.
In my opinion, if undertow is working with a facade for the session, it must take care that this facade does not reference any additional data after the request is finished.

fl4via:
The mentioned PR was added as part of this Jira:
https://issues.redhat.com/browse/UNDERTOW-1573
So, basically, the access logs were not logging attributes. We may study a solution where the attributes are cleaned after logging, I'll verify that

fl4via:
@evernat  Please view UNDERTOW-1722 for solving your issue. I'll also be studying and discussing with the team the possibility of removing the references to the request from the session without a weak reference as part of UNDERTOW-1723:
https://issues.redhat.com/projects/UNDERTOW/issues/UNDERTOW-1722
https://issues.redhat.com/projects/UNDERTOW/issues/UNDERTOW-1723
UNDERTOW-1722 will be included in 2.1.2.Final, probably you won't have problems with javamelody if you use that version (to be released in the next few hours)
UNDERTOW-1723 is now for evaluation and might be included in 2.1.3.Final if I decide to go with it.
I'm now closing this PR, feel free to open a new one if the question arises again with 2.1.2.Final.

evernat:
thanks @fl4via

PR #864, opened 2020-03-10
UNDERTOW-1661: provide a non-throwing api to add completion listeners

carterkozak:
This adds `HttpServerExchange.tryAddExchangeCompleteListener` to
attempt to add a listener with feedback describing whether or not
it was successfully added. This provides an enrgonomic API to
avoid needing prior knowledge of multiple methods required to
add an exchange complete listener without throwing.

### DIFF
--- a/core/src/main/java/io/undertow/server/HttpServerExchange.java
+++ b/core/src/main/java/io/undertow/server/HttpServerExchange.java
@@ -935,10 +935,33 @@ public final class HttpServerExchange extends AbstractAttachable {
         return this;
     }
 
-
+    /**
+     * Adds an  {@link ExchangeCompletionListener} to this {@link HttpServerExchange}. This method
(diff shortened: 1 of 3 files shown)


fl4via:
@carterkozak  Thanks for your PR. The best approach for this is under discussion, probably with 2.1.0.Final target, as the previous PR is needed internally and will be merged to 2.0.30.Final. I'll keep this thread posted.

carterkozak:
@fl4via regarding new APIs on HttpServerExchange, I'm curious why 'public hasListenersBeenInvoked' is okay, but 'tryAddExchangeCompleteListener' is not. If the former is added, I think this should be closed to avoid supporting too many ways to do the same thing.

An alternative for the other PR is to surround the failing invocation with try/catch IllegalStateException to avoid adding api surface area until we have a better sense which approach we want to use.

fl4via:
We are fixing this with PR #867

carterkozak:
Even better, thanks!

PR #866, opened 2020-03-10
UNDERTOW-1700: Prevent resource leaks when exchanges aren't closed cleanly

carterkozak:
This does not solve the linked ticket, but prevents the described
scenario from causing additional problems.

While the modified blocks are commented suggesting that they
should never be called, UNDERTOW-1664 describes when they occur
with a reproducer.


fl4via:
This is ok to test

PR #872, opened 2020-03-24
UNDERTOW-1677 cross context request issue with shared session managers

stuartwdouglas:


fl4via:
Jira: https://issues.redhat.com/browse/UNDERTOW-1677

fl4via:
All CI failures are unrelated... merging

ksiczek:
Thank you @stuartwdouglas for taking care. We impatiently wait for a fix so I had a look at your changes. I do not fully understand them but I noticed that you had changed and tested an InMemorySessionManager while our stack trace leads to DistributableSessionManager. Was it intentional? Is the InMemorySessionManager used by DistributableSessionManager somehow internally? We do not use <distributable /> directive and InMemorySessionManager seems a better choice for us but we could not switch to it anyhow.

stuartwdouglas:
DistributableSessionManager is part of WildFly, it can't be used to test in the Undertow repo. This test emulates the WildFly issue by using the same session manager for each context (the issue is caused by the use of shared session managers, not because it is distributable).

r-tock:
We are seeing this illegal state exceptions sporadically after we pushed 2.1.0. it has happened 4/1M requests, it's annoying but small enough not to warrant a rollback yet.

What's the best way to figure out what we are doing wrong?

PR #873, opened 2020-03-30
[UNDERTOW-1678] Some tests do not respect value of 'default.server.ad…

jstourac:
…dress' property

Use hostname specified via 'default.server.address'.

Currently there is a bunch of tests that expect that server hostname
is always 'localhost'. This change modifies those so that they respect
actual value in 'default.server.address' property which is respected
by io.undertow.testutils.DefaultServer#getHostAddress().

https://issues.redhat.com/browse/UNDERTOW-1678

### DIFF
--- a/core/src/test/java/io/undertow/server/HttpServerExchangeTestCase.java
+++ b/core/src/test/java/io/undertow/server/HttpServerExchangeTestCase.java
@@ -65,11 +65,11 @@ public class HttpServerExchangeTestCase {
             HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/somepath");
             HttpResponse result = client.execute(get);
             Assert.assertEquals(StatusCodes.OK, result.getStatusLine().getStatusCode());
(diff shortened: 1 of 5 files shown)


jstourac:
retest this please

jstourac:
Well, looking in the test results, I think that failures are not related with changes in this MR... it looks more like an issue with build executors... e.g. I can see following messages in the build log:
```
 Caused by: java.io.FileNotFoundException: /store/work/tc-work/b455211ade1ff869/core/target/classes/META-INF/services/io.undertow.predicate.PredicateBuilder (Too many open files)
```

fl4via:
@jstourac  we do have that recurring failure and it has to be investigated at some point. But,nope, it is not related to this PR.

jstourac:
Updated second commit by reverting the 'assertion' changes. Hope it is okay now. Thanks, @fl4via!

fl4via:
I'm merging, @jstourac , the failures we are seeing are not in the tests you updated.

PR #875, opened 2020-04-09
UNDERTOW-1701: Fix race condition in GracefulShutdownHandler

carterkozak:
1. Thread T1 calls decrementRequests. Shutdown is false, so this thread will follow the 'else' path in preparation for decrementAndGet
2. Thread T2 calls shutdown. Increments from 1 to 2, sets shutdown=true, and calls decrementRequests where decrementAndGet results in 1
3. Thread T1 completes the decrementRequests invocation reducing the active request count to zero in the else path without calling shutdownComplete.


fl4via:
This is ok to test

PR #876, opened 2020-04-16
UNDERTOW-1683 Don't throw an exception if isReady is called before re…

stuartwdouglas:
…quest returns

Jira: https://issues.redhat.com/browse/UNDERTOW-1683

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/spec/AsyncContextImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/AsyncContextImpl.java
@@ -145,6 +145,10 @@ public class AsyncContextImpl implements AsyncContext {
                 servletResponse instanceof HttpServletResponseImpl;
     }
 
+    public boolean isInitialRequestDone() {
+        return initialRequestDone;
+    }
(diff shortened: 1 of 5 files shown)


servlet/src/main/java/io/undertow/servlet/spec/AsyncContextImpl.java
```diff
@@ -145,6 +145,14 @@ public boolean hasOriginalRequestAndResponse() {
                 servletResponse instanceof HttpServletResponseImpl;
     }
 
+    public boolean isInitialRequestDone() {
+        return initialRequestDone;
+    }
+
+    public void setInitialRequestDone(boolean initialRequestDone) {
```
bstansberry, line 152: @stuartwdouglas  Nothing calls this.

stuartwdouglas: oops, I generated the setter by accident.

fl4via:
I'm merging, given that the failures seem unrelated.

PR #883, opened 2020-04-22
[UNDERTOW-1703] Checking isSymbolicLink should be in doPrivileged block

gaol:
Issue: https://issues.redhat.com/browse/UNDERTOW-1703

This is the PR to `2.0.x` branch, cherry-picked from https://github.com/undertow-io/undertow/pull/882

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/resource/SecurityActions.java
+++ b/core/src/main/java/io/undertow/server/handlers/resource/SecurityActions.java
@@ -1,0 +1,40 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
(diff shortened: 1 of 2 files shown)


gaol:
@fl4via @stuartwdouglas Would you please review? Thanks

core/src/main/java/io/undertow/server/handlers/resource/SecurityActions.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
fl4via, line 3: We must change this to 2020

gaol: Thanks, Done :)

PR #884, opened 2020-04-22
[Undertow 1702] SameSiteCookieHandler can throw NPE if request doesn't contain user-agent header

jstourac:
https://issues.redhat.com/browse/UNDERTOW-1702

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/SameSiteCookieHandler.java
@@ -74,8 +74,9 @@ public class SameSiteCookieHandler implements HttpHandler {
             exchange.addResponseCommitListener(new ResponseCommitListener() {
                 @Override
                 public void beforeCommit(HttpServerExchange exchange) {
-                    // Check user-agents and skip sending "SameSite=None" for incompatible user-agents
-                    if (enableClientChecker && !SameSiteNoneIncompatibleClientChecker.shouldSendSameSiteNone(exchange.getRequestHeaders().getFirst(Headers.USER_AGENT))) {
+                    // If user-agent is available check it and skip sending "SameSite=None" for incompatible user-agents
+                    String userAgent = exchange.getRequestHeaders().getFirst(Headers.USER_AGENT);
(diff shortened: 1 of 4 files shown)


msfm:
Sorry that I forgot to add a null check in SameSiteNoneIncompatibleClientChecker.

This PR looks good, but it could be better to add a null check in SameSiteNoneIncompatibleClientChecker. As the root cause exists there.

```diff
diff --git a/core/src/main/java/io/undertow/util/SameSiteNoneIncompatibleClientChecker.java b/core/src/main/java/io/undertow/util/SameSiteNoneIncompatibleClientChecker.java
index 6bcf775..8c91f07 100644
--- a/core/src/main/java/io/undertow/util/SameSiteNoneIncompatibleClientChecker.java
+++ b/core/src/main/java/io/undertow/util/SameSiteNoneIncompatibleClientChecker.java
@@ -65,6 +65,9 @@ public final class SameSiteNoneIncompatibleClientChecker {
 
     // browsers known to be incompatible.
     public static boolean isSameSiteNoneIncompatible(String useragent) {
+        if (useragent == null || useragent.isEmpty()) {
+            return false;
+        }
         return hasWebKitSameSiteBug(useragent) ||
             dropsUnrecognizedSameSiteCookies(useragent);
     }
```

jstourac:
Hi @msfm, thanks for review. You are right!

Looking at the code and tests in `SameSiteNoneIncompatibleClientCheckerTestCase`, there are also `hasWebKitSameSiteBug` and `dropsUnrecognizedSameSiteCookies` methods tested separately. This would mean to put null check into these two methods to make tests happy. 

I'm thinking of removing two tests for those two methods and also make all methods in the `SameSiteNoneIncompatibleClientChecker` private except the two that are marked as public alreday. Currently there is `package-scope` used for rest of the methods... I think it is not necessary. What do you think?

msfm:
Hi @jstourac. Sorry to late response.

> I'm thinking of removing two tests for those two methods and also make all methods in the SameSiteNoneIncompatibleClientChecker private except the two that are marked as public already. Currently there is package-scope used for rest of the methods... I think it is not necessary. What do you think?

I agree with your thought. We can test just public methods only and can change other package-scope methods to private.

jstourac:
@msfm thank you, I've just updated this one

jstourac:
CI test failures don't seem to be related to changes in this MR.

PR #886, opened 2020-05-12
UNDERTOW-1419 At InMemorySessionMemoryManager bump timeout only when …

fl4via:
…we create the session and when request is done.

Jira: https://issues.redhat.com/browse/UNDERTOW-1419

### DIFF
--- a/core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
+++ b/core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
@@ -476,6 +476,7 @@ public class InMemorySessionManager implements SessionManager, SessionManagerSta
             if(existing != null) {
                 lastAccessed = existing;
             }
+            bumpTimeout();
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


fabiobrz:
Hi @fl4via, I am missing something for sure here but could you please let me know where there is some known reason for CI checks failure? I am sure that I am missing something but I am asking anyway since I am testing a patch which actually includes this change. Thanks in advance.

maniek45:
Hi,

We used in our code setMaxInactiveInterval() method (indirectly by using HttpSession) to dynamically switch on/off timeouts. It worked perfectly on Wildfly 19.0.0.Final (which contains undertow-core v2.0.30), but suddenly stopped working when app has been migrated to spring-boot v.2.3.2.RELEASE (with embedded undertow-core v2.1.3.Final). Unfortunately our code invokes setMaxInactiveInterval() method outside of http request/response lifecycle. My question is if it is a chance to bring back the old functionality (with bumpTimeout() into setMaxInactiveInterval() or we should implement our own http session timeout mechanism? 

Thanks,
Marcin.

huanghwh:
Is it possible to bring back I bumpTimeout()  in setMaxInactiveInterval()?

PR #897, opened 2020-05-28
UNDERTOW-1720 fix potential race in senders

stuartwdouglas:


core/src/main/java/io/undertow/io/AsyncSenderImpl.java
```diff
@@ -408,11 +429,15 @@ private void invokeOnComplete() {
...
-            inCallback = true;
+            writeThread = null;
+            inCallback = Thread.currentThread();
             try {
                 callback.onComplete(exchange, this);
             } finally {
-                inCallback = false;
+                inCallback = null;
+            }
+            if (Thread.currentThread() != writeThread) {
```
bstansberry, line 439: L432 means this always evaluates as true.

stuartwdouglas: onComplete can do another write.

core/src/main/java/io/undertow/io/BlockingSenderImpl.java
```diff
@@ -292,11 +302,15 @@ private void invokeOnComplete(final IoCallback callback) {
             } else if (file != null) {
                 performTransfer(file, queuedCallback);
             }
-            inCall = true;
+            sendThread = null;
+            inCall = Thread.currentThread();
             try {
-                queuedCallback.onComplete(exchange, this);
+                callback.onComplete(exchange, this);
```
fl4via, line 308: @stuartwdouglas  Since we are on next != null and have already invoked callback, shouldn't t this line be kept as queuedCallback?

servlet/src/main/java/io/undertow/servlet/core/BlockingWriterSenderImpl.java
```diff
@@ -221,11 +231,15 @@ private void invokeOnComplete(final IoCallback callback) {
             if (writer.checkError()) {
                 queuedCallback.onException(exchange, this, new IOException());
             } else {
-                inCall = true;
+                sendThread = null;
+                inCall = Thread.currentThread();
                 try {
-                    queuedCallback.onComplete(exchange, this);
+                    callback.onComplete(exchange, this);
```
fl4via, line 237: @stuartwdouglas  The same here, aren't we supposed to process queuedCallback in this block?

core/src/main/java/io/undertow/io/AsyncSenderImpl.java
bstansberry: Sorry for the noise!

core/src/main/java/io/undertow/io/BlockingSenderImpl.java
stuartwdouglas: yes, that is a copy/paste error

core/src/main/java/io/undertow/io/AsyncSenderImpl.java
fl4via: no need to be sorry, a review is always welcome!

PR #898, opened 2020-05-28
[UNDERTOW-1719] Handling of path parameters can set incorrect requestURI/requestPath

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-1719

A new `parsePathComplete` method has been created to not repeat the code at the end of the three situations that completes a path parsing  (common path, path with parameters and start of query string). This fixes the two issues and a little corner case.

Some more tests added and now the `isHostIncludedInRequestURI` is also checked in the majority of them.

PR for master branch.


fl4via:
PR for 2.0.x branch: #899

PR #904, opened 2020-06-03
[UNDERTOW-1726] Disable the JDK9AlpnProvider from being used by defau…

jamezp:
…lt with Java 8 251 and higher. Add a system property to disable this behavior and use the JDK9AlpnProvider if the SSLEngine implementation being used is known to have the new methods implemented.

https://issues.redhat.com/browse/UNDERTOW-1726


fl4via:
This is ok to test

PR #906, opened 2020-06-17
Fix PathResourceManagerTestCase on later JDKs

stuartwdouglas:


fl4via:
@stuartwdouglas  I'm closing this PR because, after rebasing locally, I see that all fixes it contains are already included in master

PR #908, opened 2020-07-08
UNDERTOW-1747 - Provide a mechanism for a predicate in return a servlet error page

bdw429s:
Using the predicate language in Undertow, there is no way to return the error pages configured in the servlet from the handler chain that runs first in the XNIO worker thread.  For example, if the response-code() handler is used, the exchange is simply ended and no response body is sent which is not a very useful behavior.

A solution provided by Stuart on the mailing list was to set the response code but not end the exchange.  Then configure the servlet handler to check for a status code above 399 and call the servlet sendError() routine.

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/handlers/SendErrorPageHandler.java
+++ b/servlet/src/main/java/io/undertow/servlet/handlers/SendErrorPageHandler.java
@@ -46,7 +46,7 @@ public class SendErrorPageHandler implements HttpHandler {
         ServletRequestContext src = exchange.getAttachment(ServletRequestContext.ATTACHMENT_KEY);
 
         // If the servlet is available and the status code has been set to an error, return the error page
-        if( src != null && exchange.getStatusCode() > 399 && exchange.getResponseContentLength() == -1 ) {
+        if( src != null && exchange.getStatusCode() > 399 && !exchange.isResponseStarted() ) {
             ((HttpServletResponse)src.getServletResponse()).sendError(exchange.getStatusCode());
         } else {
             next.handleRequest(exchange);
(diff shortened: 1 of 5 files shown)


bdw429s:
This is for ticket https://issues.redhat.com/browse/UNDERTOW-1747

This pull lets you have a predicate like so:
```
path(/box.json)->set-error(404)
```
And hitting that path will send back the servlet's configured 404 error page.

bdw429s:
@stuartwdouglas @fl4via This is for ticket https://issues.redhat.com/browse/UNDERTOW-1747

This pull lets you have a predicate like so:
```
path(/box.json)->set-error(404)
```
And hitting that path will send back the servlet's configured 404 error page.

servlet/src/main/java/io/undertow/servlet/handlers/SendErrorPageHandler.java
```diff
@@ -0,0 +1,56 @@
...
+    public SendErrorPageHandler(HttpHandler next) {
+        this.next = next;
+    }
+
+    @Override
+    public void handleRequest(final HttpServerExchange exchange) throws Exception {
+        ServletRequestContext src = exchange.getAttachment(ServletRequestContext.ATTACHMENT_KEY);
+
+        // If the servlet is available and the status code has been set to an error, return the error page
+        if( src != null && exchange.getStatusCode() > 399 && exchange.getResponseContentLength() == -1 ) {
```
stuartwdouglas, line 49: This should be !exchange.isResponseStarted(), the response content length does not really tell you anything, a lot of requests use chunked encoding which will not set it.

bdw429s: Done.

bdw429s:
Thanks @stuartwdouglas !

fl4via:
I'll probably move this PR to master. @bdw429s  this PR and others that you submitted will be available in a release I'll be doing later this week. I'll let you know when it is available.

fl4via:
@bdw429s the master PR is #948

bdw429s:
Thanks @fl4via !

PR #916, opened 2020-07-30
[WIP] next round of fixes

stuartwdouglas:
These will likely need to be broken up into separate PR's.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,17 +42,17 @@ jobs:
         name: surefire-reports-build
         path: '**/surefire-reports/*.txt'
   test-matrix:
-    name: JDK ${{ matrix.jdk }} - ${{ matrix.module }} - ${{ matrix.os }} - ${{ matrix.protocol }} - ${{ matrix.ip }}
+    name: ${{ matrix.module }} ${{ matrix.protocol }} - ${{ matrix.os }} - JDK${{ matrix.jdk }} - ${{ matrix.ip }}
     runs-on: ${{ matrix.os }}
     needs: build-all
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
         module: [core, servlet]
+        protocol: ['', '''-Dtest.proxy=true''', '''-Dtest.https=true''', '''-Dtest.ajp=true''', '''-Dtest.h2=true''', '''-Dtest.h2c=true''', '''-Dtest.h2c-upgrade=true''']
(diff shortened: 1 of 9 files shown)


luck3y:
For https://github.com/undertow-io/undertow/pull/916/checks?check_run_id=927351696#step:10:617 I tried all sorts of tricks yesterday to raise the shell limit (normally just update the limits and reboot but thats not an option here.) The defualt value in the runners is 65k, though, which is a lot higher than the normal default, so its possibly a leak in a prior test.
Failing the leak theory, we can probably do something along the lines of sudo su -i runner <maven> to get a new shell with new limits and run the test, but I'm not sure if that'll actually work yet ;)

stuartwdouglas:
It must be a leak somehow. I have run the test suite locally and looked as lsof and only see low numbers.

luck3y:
Not having a shell on here makes this somewhat more difficult, maybe I can & the maven command and then just run netstat in a loop to count, I'll give that a go later, might be generally useful anyway.

luck3y:
FWIW, comparing to a local run, io.undertow.client.http.AjpClientTestCase-output.txt is 153mb and full of exceptions running on the github host vs locally being 427b, so I'll dig into that a bit.

luck3y:
FYI I've rebased on this and moved to ubuntu-20.04 here: https://github.com/luck3y/undertow/runs/930595368?check_suite_focus=true

It also installs openssl-1.1.1 on windows. Ubuntu-latest is 18.04, and the openssl is pretty old with no update available.

This helped a bit when I was trying to get xnio CI to work, so its probably worth a try here too.

luck3y:
https://github.com/luck3y/undertow/runs/930847307?check_suite_focus=true

This one seems to be stable now after expanding the ephemeral port range and disabling tw_reuse on the loopback interface. Its likely some of the windows ones have a similar mitigation (not sure how to change the windows ephemeral port range yet.

fl4via:
@stuartwdouglas I am reviewing this PR to see what has not been merged yet. Did you intend to split it into so many checks for good, or just to help investigation of the failures?

stuartwdouglas:
A bit of both to be honest. Given the current state of the test suite I think it gives a lot better visibility into where the problems are, and should actually help a lot with stability.

fl4via:
@stuartwdouglas  Okay, I'll incorporate that change then. I'm currently tagging 2.2.0.Final, and then I'll get this PR ready for merging.

ropalka:
This PR is WIP and  is out of date. Closing.

PR #917, opened 2020-07-30
[UNDERTOW-1762] Error page for custom exception-type is not is not displayed in jsp development mode

aogburn:
https://issues.redhat.com/browse/UNDERTOW-1762
2.0.x PR: #944

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/core/ErrorPages.java
+++ b/servlet/src/main/java/io/undertow/servlet/core/ErrorPages.java
@@ -61,10 +61,13 @@ public class ErrorPages {
         if (location == null && exception instanceof ServletException) {
             Throwable rootCause = ((ServletException) exception).getRootCause();
             //Iterate through any nested JasperException in case it is in JSP development mode
-            while (rootCause != null && rootCause instanceof ServletException) {
+            while (rootCause != null && rootCause instanceof ServletException && location == null) {
+                for (Class c = rootCause.getClass(); c != null && location == null; c = c.getSuperclass()) {
+                    location = exceptionMappings.get(c);
+                }
                 rootCause = ((ServletException) rootCause).getRootCause();
             }
-            if (rootCause != null) {
+            if (rootCause != null && location == null) {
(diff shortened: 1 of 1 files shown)


spyrkob:
This causes a failing TCK test: https://github.com/eclipse-ee4j/jakartaee-tck/blob/8.0.2/src/com/sun/ts/tests/servlet/spec/errorpage1/URLClient.java#L101 The problem is when a custom exception extending ServletException is wrapped in ServletException and that custom exception is used in mapping.

Maybe the mapping check should be performed for each level of unwrap?

aogburn:
Ah, yes, made an adjustment to fix that.

servlet/src/main/java/io/undertow/servlet/core/ErrorPages.java
```diff
@@ -61,10 +61,13 @@ public String getErrorLocation(final Throwable exception) {
...
-            while (rootCause != null && rootCause instanceof ServletException) {
+            while (rootCause != null && rootCause instanceof ServletException && location == null) {
+                for (Class c = rootCause.getClass(); c != null && location == null; c = c.getSuperclass()) {
+                    location = exceptionMappings.get(c);
+                }
                 rootCause = ((ServletException) rootCause).getRootCause();
             }
-            if (rootCause != null) {
+            if (rootCause != null && location == null) {
                 for (Class c = rootCause.getClass(); c != null && location == null; c = c.getSuperclass()) {
```
fl4via, line 71: we don't need to check for location equals null any longer because of the change in the previous line

fl4via: @aogburn  I'll take the freedom to update the PR myself because I'm working on releasing and tagging Undertow with this fix. Thanks for your PR!

fl4via: @aogburn never mind, my bad, this code is actually correct and does not need to be updated

PR #924, opened 2020-08-14
[UNDERTOW-1782][WFLY-13748]: fix for Contains non-LDH ASCII during ejb remote call in …

JulienLC:
Hi,

This pull request contains a fix for the  issue  described here "https://issues.redhat.com/browse/WFLY-13748".
"Contains non-LDH ASCII" characters during remote EJB SSL call in IPV6 network.

The root cause of the issue was that SNIHostName cannot handle String with ":" 
characters (it thows illegalArgumentException) . destination.getHostString() is returning a IPV6 adress containing ":" characters.

The suggested fix is to use the method address.getHostName() which is always returning hostname value.
This fix was tested successfully in our IPV6 kubernetes environement.

regards

### DIFF
--- a/core/src/main/java/io/undertow/protocols/ssl/UndertowXnioSsl.java
+++ b/core/src/main/java/io/undertow/protocols/ssl/UndertowXnioSsl.java
@@ -18,9 +18,9 @@
 
 package io.undertow.protocols.ssl;
 
-import static org.xnio.IoUtils.safeClose;
-
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


fl4via (changes requested):
The fix LGTM. I'm creating an Undertow Jira and updating the PR with that ticket number before merging. Thanks!

fl4via:
Undertow Jira: https://issues.redhat.com/browse/UNDERTOW-1782

PR #948, opened 2020-08-24
UNDERTOW-1747 - Provide a mechanism for a predicate in return a servlet error page

fl4via:
Jira: https://issues.redhat.com/browse/UNDERTOW-1747
2.0.x PR (original PR, reverted to prevent new feature from being added to maintenance branch): #908


core/src/main/java/io/undertow/server/handlers/SetErrorHandler.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
fl4via, line 3: The copyright should be 2020. I'll update the PR before merging it

PR #949, opened 2020-09-04
UNDERTOW-1773 to avoid potential leak, make sure the iteration continue in case of exception.

soul2zimate:
issue: https://issues.redhat.com/browse/UNDERTOW-1773
UNDERTOW-1773 to avoid potential leak, make sure the iteration continue in case of exception.

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/core/ApplicationListeners.java
+++ b/servlet/src/main/java/io/undertow/servlet/core/ApplicationListeners.java
@@ -212,7 +212,12 @@ public class ApplicationListeners implements Lifecycle {
         }
         final ServletContextAttributeEvent sre = new ServletContextAttributeEvent(servletContext, name, value);
         for (int i = 0; i < servletContextAttributeListeners.length; ++i) {
-            this.<ServletContextAttributeListener>get(servletContextAttributeListeners[i]).attributeAdded(sre);
+            ManagedListener listener = servletContextAttributeListeners[i];
+            try {
+                this.<ServletContextAttributeListener> get(listener).attributeAdded(sre);
+            } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


servlet/src/main/java/io/undertow/servlet/core/ApplicationListeners.java
```diff
@@ -325,7 +326,12 @@ public void sessionDestroyed(final HttpSession session) {
         final HttpSessionEvent sre = new HttpSessionEvent(session);
         for (int i = httpSessionListeners.length - 1; i >= 0; --i) {
             ManagedListener listener = httpSessionListeners[i];
-            this.<HttpSessionListener>get(listener).sessionDestroyed(sre);
+            try {
+                this.<HttpSessionListener> get(listener).sessionDestroyed(sre);
+            } catch (Exception e) {
+                // UNDERTOW-1773 to avoid leak, make sure the iteration continue in case of exception.
+                UndertowLogger.ROOT_LOGGER.debugf(e, "Exception happens in sessionDestroyed for session %s ", session);
```
stuartwdouglas, line 333: This should really be at error level.

soul2zimate: Thank you, I have updated the log level to ERROR

fl4via:
As has been discussed in the JBEAP Jira, we should apply the same fix to other listener notifications as well.

soul2zimate:
> As has been discussed in the JBEAP Jira, we should apply the same fix to other listener notifications as well.

OK, I will apply them to all other listener notifications.

soul2zimate:
I have updated the commit to cover all other methods with listener invocations

fl4via (approved):

PR #975, opened 2020-10-06
[UNDERTOW-1795] Always sort cookies

ropalka:
https://issues.redhat.com/browse/UNDERTOW-1795

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/CookieImpl.java
+++ b/core/src/main/java/io/undertow/server/handlers/CookieImpl.java
@@ -188,8 +188,8 @@ public class CookieImpl implements Cookie {
     @Override
     public final boolean equals(final Object other) {
         if (other == this) return true;
-        if (!(other instanceof CookieImpl)) return false;
-        final CookieImpl o = (CookieImpl) other;
+        if (!(other instanceof Cookie)) return false;
+        final Cookie o = (Cookie) other;
         // compare names
         if (getName() == null && o.getName() != null) return false;
         if (getName() != null && !getName().equals(o.getName())) return false;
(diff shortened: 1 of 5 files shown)


scottmarlow (commented):
:+1:

fl4via (approved):

core/src/main/java/io/undertow/server/handlers/CookieImpl.java
```diff
@@ -203,6 +203,33 @@ public final boolean equals(final Object other) {
         return true;
     }
 
+    @Override
+    public final int compareTo(final Object other) {
```
bstansberry, line 207: @ropalka @fl4via 

Should this be the default impl in the interface? AFAICT it's entirely using calls to interface methods, so it could be a default impl.  Doing it that way avoids making an incompatible change in a public interface.

carterkozak: From the peanut gallery:

Thoughts on using a constant `Comparator<Cookie>` to provide to the TreeSet to avoid issues with incorrect custom cookie implementations?

bstansberry: Note that even though I commented here but don't block waiting for a PR approval from me.

bstansberry: @carterkozak I'll defer to @ropalka on that one. Seems reasonable.

ropalka: That was my initial implementation @bstansberry but I didn't like
default methods cannot be final. So I moved it to cookie impl classes instead.

But thinking about it again it introduced good-for-nothing code duplicities so
I reworked it back again as you propose. Additionally we want to forbid
subclasses to break the contract (same applies to equals() & hashCode() methods)
so we override comparteTo() method in subclasses to delegate to default method impl.

ropalka: Thanks @carterkozak for your proposal. That is also alternative.
I prefer compareTo() to be co-located in classes where
equals(), hashCode() & toString() methods are declared.

fl4via (approved):

PR #976, opened 2020-10-07
[UNDERTOW-1796] [UNDERTOW-1800] Default the io.undertow.protocols.alpn.jdk8 to true and change the order of preferred providers

jamezp:
JIRA's:

- https://issues.redhat.com/browse/UNDERTOW-1796
- https://issues.redhat.com/browse/UNDERTOW-1800

master: #977 
2.0.x PR: #982

### DIFF
--- a/core/src/main/java/io/undertow/protocols/alpn/JDK9AlpnProvider.java
+++ b/core/src/main/java/io/undertow/protocols/alpn/JDK9AlpnProvider.java
@@ -43,9 +43,6 @@ public class JDK9AlpnProvider implements ALPNProvider {
     private static final String JDK8_SUPPORT_PROPERTY = "io.undertow.protocols.alpn.jdk8";
 
     static {
-        // This property must be checked outside of the privileged action as the user should explicitly provide read
-        // access to it. A value of true is the only supported value.
-        final boolean addSupportIfExists = Boolean.getBoolean(JDK8_SUPPORT_PROPERTY);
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/protocols/alpn/JDK9AlpnProvider.java
```diff
@@ -64,6 +61,8 @@ public JDK9ALPNMethods run() {
                     // methods would exist and could result in issues. These methods can still be used by providing the
                     // io.undertow.protocols.alpn.jdk8=true system property and support for Java 8 known in the
                     // SSLEngine implementation being provided.
+                    final String value = System.getProperty(JDK8_SUPPORT_PROPERTY);
```
ropalka, line 64: Easier to use Boolean.getBoolean(String propertyName) instead.

ropalka: Sorry, just realized you want to default to true. Please ignore my comment.

jamezp: That wouldn't work because we want to default to `true` so if it's not defined, e.g. `null`, we need it to be true.

jstourac:
@fl4via After the merge - please update `Fix versions` field of tracking issues (already marked as resolved) adding a correct value, since I've just removed incorrect `2.1.5.Final` record from there. Thank you.

PR #983, opened 2020-10-26
[UNDERTOW-1821] Path template matched parameters should keep order

seregamorph:
Jira: https://issues.redhat.com/browse/UNDERTOW-1821

Replace `HashMap` to `LinkedHashMap` in `PathTemplateMatcher`. Motivation: ordered matched paths look more natural e.g. in debugging:
Screen Shot 2020-10-26 at 10 18 51

Also in my case I needed to write a simple custom matcher: not Map (expected) equals Map (actual), but a List of matched values (cannot use `Set`, because values can be repeated).

Reference implementation: spring-core `org.springframework.util.AntPathMatcher` keeps original order:
```java
AntPathMatcher antPathMatcher = new AntPathMatcher();
Map<String, String> pathVariables = antPathMatcher.extractUriTemplateVariables(
        "/{context}/{version}/{controller}/{id}",
        "/api/v3/users/1"
);
// prints "{context=api, version=v3, controller=users, id=1}"
```

Cons: in theory, updated impl will take more nanoseconds in runtime and more nanobytes in memory, but I believe if there was an intention to micro-optimise it, these should be a custom Map implementation (like Netty `HttpHeaders`)

### DIFF
--- a/core/src/main/java/io/undertow/util/PathTemplateMatcher.java
+++ b/core/src/main/java/io/undertow/util/PathTemplateMatcher.java
@@ -21,9 +21,9 @@ package io.undertow.util;
 import io.undertow.UndertowMessages;
 
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


seregamorph:
@stuartwdouglas what do you think?

fl4via (approved):

fl4via:
@seregamorph  we are approving this change. The advantage of keeping parameters in order outweights the added overhead, specially because hte matcher's map is small and is not used every where in the code.

seregamorph:
Thanks @fl4via

PR #987, opened 2020-11-13
Attempt to run CI with java 15

carterkozak:
This is prompted by https://issues.redhat.com/browse/UNDERTOW-1791
Hoping to illustrate failures on java 15 and validate that fixes resolve the failures.

PR #988, opened 2020-11-13
UNDERTOW-1791: Migrate from getPeerCertificateChain to getPeerCertificates

carterkozak:
Jira: https://issues.redhat.com/browse/UNDERTOW-1791

Java 15 runtimes default jsse SSLSession implementation throws an
UnsupportedOperationException when getPeerCertificateChain is called.
https://www.oracle.com/java/technologies/javase/15-relnote-issues.html#JDK-8241039

In this PR I have replaced uses of getPeerCertificateChain with getPeerCertificates, however I have not provided an implementation of sslsessioninfo.getPeerCertificateChain which delegates to getPeerCertificates and converts to the javax variant. It's unclear if that would be desirable.
It appears unlikely that there are cases in which getPeerCertificates produces different results, but it's possible a SSLSession implementation is incorrect in a way that causes this change to introduce failures. We could flag this behavior behind a property.

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/handlers/security/SSLInformationAssociationHandler.java
+++ b/servlet/src/main/java/io/undertow/servlet/handlers/security/SSLInformationAssociationHandler.java
@@ -18,7 +18,7 @@
 
 package io.undertow.servlet.handlers.security;
 
-import java.io.ByteArrayInputStream;
+import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import javax.servlet.ServletRequest;
 
@@ -86,8 +86,6 @@ public class SSLInformationAssociationHandler implements HttpHandler {
 
     /**
      * Return the chain of X509 certificates used to negotiate the SSL Session.
-     * <p>
-     * We convert JSSE's javax.security.cert.X509Certificate[]  to servlet's  java.security.cert.X509Certificate[]
      *
      * @param session the   javax.net.ssl.SSLSession to use as the source of the cert chain.
      * @return the chain of java.security.cert.X509Certificates used to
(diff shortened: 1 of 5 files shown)


carterkozak:
I've cherry-picked the commit from #987 to run tests using jdk15 in addition to 8 and 11.

servlet/src/main/java/io/undertow/servlet/handlers/security/SSLInformationAssociationHandler.java
```diff
@@ -96,23 +94,31 @@ public static int getKeyLength(String cipherSuite) {
...
-            return javaCerts;
+            if (x509Certs == 0) {
+                return null;
+            }
+            if (x509Certs == javaCerts.length) {
+                return (X509Certificate[]) javaCerts;
+            }
+            X509Certificate[] results = new X509Certificate[x509Certs];
+            for (int i = 0; i < x509Certs; i++) {
+                results[i] = (X509Certificate) javaCerts[i];
```
ropalka, line 115: You cannot be sure certificate on index i is instance of X509Certificate.
The loop should be using instanceof check.

ropalka (changes requested):

servlet/src/main/java/io/undertow/servlet/handlers/security/SSLInformationAssociationHandler.java
carterkozak: oh geez, I had a dumb... fix incoming

ropalka (approved):

fl4via:
this PR introduced a regression in the SSL tests and will require some investigation before being merged

carterkozak:
I've fixed the failure -- sorry about that. Hopefully it's not too late for 2.2.3.Final.

fl4via:
thanks, there is still time for 2.2.3.Final @carterkozak !

PR #990, opened 2020-11-21
[UNDERTOW-1810] Dynamically determine the major servlet version at ru…

bstansberry:
…ntime based on the name of the ServletContextListener package.

https://issues.redhat.com/browse/UNDERTOW-1810

This is a hack so I won't be offended if it's rejected.

The obvious alternative is to produce a native jakarta.* variant of undertow but that's a substantial undertaking and it would be nice to be able to use Undertow in a compliant EE 9 impl soon.

Another alternative is to provide a mechanism for the container using undertow-servlet to control what undertow-servlet reports (system property, some interface loadable via ServiceLoader, ...). But that's harder for WildFly and seems a bit more fragile. Such a hook also becomes a form of API.

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
+++ b/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
@@ -63,15 +63,27 @@ import io.undertow.util.ImmediateAuthenticationMechanismFactory;
  */
 public class DeploymentInfo implements Cloneable {
 
+    private static final int DEFAULT_MAJOR_VERSION;
+
+    static {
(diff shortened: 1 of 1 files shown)


ropalka (approved):
LGTM

fl4via (approved):

fl4via:
@bstansberry  It is a hack indeed, but I see where it is coming from. Eventually we will have to fork to a jakarta ee9.1 implementation and then if needed we will revert it. (the next spec version will add samesite cookie support)

fl4via:
@bstansberry  will this be needed for JBEAP as well? I need to decide if I want to backport it to maintenance branches or not

bstansberry:
@fl4via Thanks. And no, not needed for JBEAP.

PR #1017, opened 2020-12-22
[UNDERTOW-1828] NullPointerException occurred if a servlet calls Http…

moulalis:
issue: https://issues.redhat.com/browse/UNDERTOW-1828
2.1.x PR: https://github.com/undertow-io/undertow/pull/1030
2.0.x PR: https://github.com/undertow-io/undertow/pull/1029

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/UndertowServletMessages.java
+++ b/servlet/src/main/java/io/undertow/servlet/UndertowServletMessages.java
@@ -232,4 +232,7 @@ public interface UndertowServletMessages {
 
     @Message(id = 10062, value = "No SecurityContext available")
     ServletException noSecurityContextAvailable();
+
+    @Message(id = 10063, value = "This Context has been already destroyed")
+    IllegalStateException nullDeploymentInfo();
 }
--- a/servlet/src/main/java/io/undertow/servlet/UndertowServletMessages.java
+++ b/servlet/src/main/java/io/undertow/servlet/UndertowServletMessages.java
@@ -232,7 +232,4 @@ public interface UndertowServletMessages {
 
     @Message(id = 10062, value = "No SecurityContext available")
     ServletException noSecurityContextAvailable();
-
-    @Message(id = 10063, value = "This Context has been already destroyed")
-    IllegalStateException nullDeploymentInfo();
 }
--- a/servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
@@ -207,11 +207,12 @@ public class ServletContextImpl implements ServletContext {
 
     @Override
     public String getContextPath() {
-        String contextPath = deploymentInfo.getContextPath();
-        if (contextPath.equals("/")) {
-            return "";
-        }
-        return contextPath;
+            checkDeploymentInfo();
+            String contextPath = deploymentInfo.getContextPath();
+            if (contextPath.equals("/")) {
+                return "";
+            }
+            return contextPath;
     }
 
     @Override
@@ -225,21 +226,25 @@ public class ServletContextImpl implements ServletContext {
 
     @Override
     public int getMajorVersion() {
+        checkDeploymentInfo();
         return deploymentInfo.getContainerMajorVersion();
     }
 
     @Override
     public int getMinorVersion() {
+        checkDeploymentInfo();
         return deploymentInfo.getContainerMinorVersion();
     }
 
     @Override
     public int getEffectiveMajorVersion() {
+        checkDeploymentInfo();
         return deploymentInfo.getMajorVersion();
     }
 
     @Override
     public int getEffectiveMinorVersion() {
+        checkDeploymentInfo();
         return deploymentInfo.getMinorVersion();
     }
 
@@ -260,7 +265,8 @@ public class ServletContextImpl implements ServletContext {
     public Set<String> getResourcePaths(final String path) {
         final Resource resource;
         try {
-            resource = deploymentInfo.getResourceManager().getResource(path);
+             checkDeploymentInfo();
+             resource = deploymentInfo.getResourceManager().getResource(path);
         } catch (IOException e) {
             return null;
         }
@@ -294,7 +300,8 @@ public class ServletContextImpl implements ServletContext {
         }
         Resource resource = null;
         try {
-            resource = deploymentInfo.getResourceManager().getResource(path);
+               checkDeploymentInfo();
+               resource = deploymentInfo.getResourceManager().getResource(path);
         } catch (IOException e) {
             return null;
         }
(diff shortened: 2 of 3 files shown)


msfm:
- Is it a correct behavior to treat contextPath as "/" if `deploymentInfo == null`?
- Similar NullPointerException can happen on other methods in ServletContextImpl, too.
- What about adding a check logic for `deploymentInfo`? Then, if `deploymentInfo == null`,  throwing IllegalStateException with a message like `this context has been already destroyed`?

moulalis:
> * Is it a correct behavior to treat contextPath as "/" if `deploymentInfo == null`?
> * Similar NullPointerException can happen on other methods in ServletContextImpl, too.
> * What about adding a check logic for `deploymentInfo`? Then, if `deploymentInfo == null`,  throwing IllegalStateException with a message like `this context has been already destroyed`?

@msfm 
> There was already a check  if (contextPath.equals("/")), however server stops on ctrl+c so it will remain on same context, that is the reason I have  given contextPath as "/" if `deploymentInfo == null`.
> If there are other locations then we can add it there too but dont have any test case to reproduce it aS NullpointerException.
> If you think throwing Exception is the best option thyen I will add it.

Thanks.

msfm:
> There was already a check if (contextPath.equals("/"))

I think that check exists only for the different purpose. That logic exists for the following statement described in [the API Doc of ServletContext#getContextPath()](https://docs.oracle.com/javaee/7/api/javax/servlet/ServletContext.html#getContextPath--):

> Returns:
> The context path of the web application, or **"" for the default (root) context**

---

> If there are other locations then we can add it there too but dont have any test case to reproduce it as NullpointerException.

Similar NullpointerException can happen on the following methods:

 - `context.getServerInfo()`
 - `context.getMajorVersion()`
 - `context.getMinorVersion()`
 - `context.getEffectiveMajorVersion()`
 - `context.getEffectiveMinorVersion()`
 - `context.getResourcePaths(path)`
 - `context.getResourcePath(path)`

For example, modify the reproducer which is attached in [JBEAP-20706](https://issues.redhat.com/browse/JBEAP-20706) as shown below:
```
    @Override
    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
        log.infof("# start");
        try {
            Thread.sleep(10000L);
        } catch (ignore e) { }
        ServletContext context = request.getServletContext();
        System.out.println("# context.getServerInfo() = " + context.getServerInfo());
    }
```
and follow the same steps to reproduce. 
Then, you will see similar `NullpointerException` happens at `io.undertow.servlet.spec.ServletContextImpl.getServerInfo(ServletContextImpl.java:417)`.

moulalis:
> > There was already a check if (contextPath.equals("/"))
> 
> I think that check exists only for the different purpose. That logic exists for the following statement described in [the API Doc of ServletContext#getContextPath()](https://docs.oracle.com/javaee/7/api/javax/servlet/ServletContext.html#getContextPath--):
> 
> > Returns:
> > The context path of the web application, or **"" for the default (root) context**
> 
> > If there are other locations then we can add it there too but dont have any test case to reproduce it as NullpointerException.
> 
> Similar NullpointerException can happen on the following methods:
> 
> * `context.getServerInfo()`
> * `context.getMajorVersion()`
> * `context.getMinorVersion()`
> * `context.getEffectiveMajorVersion()`
> * `context.getEffectiveMinorVersion()`
> * `context.getResourcePaths(path)`
> * `context.getResourcePath(path)`
> 
> For example, modify the reproducer which is attached in [JBEAP-20706](https://issues.redhat.com/browse/JBEAP-20706) as shown below:
> 
> ```
>     @Override
>     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
>         try {
>             Thread.sleep(10000L);
>         } catch (Exception ignore) { }
>         ServletContext context = req.getServletContext();
>         System.out.println("# context.getServerInfo() = " + context.getServerInfo());
>     }
> ```
> 
> and follow the same steps to reproduce.
> Then, you will see similar `NullpointerException` happens at `io.undertow.servlet.spec.ServletContextImpl.getServerInfo(ServletContextImpl.java:417)`.

What do you suggest now?

msfm:
My suggestion does not change. What about adding a check logic for `deploymentInfo` before using `deploymentInfo` inside `ServletContextImpl`? Then, if `deploymentInfo == null`, throwing another exception (e.g. `IllegalStateException`) instead of `NullPointerException` with a meaningful message (e.g. `this context has been already destroyed`, etc).

moulalis:
> My suggestion does not change. What about adding a check logic for `deploymentInfo` before using `deploymentInfo` inside `ServletContextImpl`? Then, if `deploymentInfo == null`, throwing another exception (e.g. `IllegalStateException`) instead of `NullPointerException` with a meaningful message (e.g. `this context has been already destroyed`, etc).

@msfm changes modified, please review.
Thanks.

fl4via:
@moulalis  I think it is best if we add a check in a single method, and we add the exception message to logger. I'll be updating your PR within the next minutes to make sure it is included in the next release. Thanks for your PR!

moulalis:
> @moulalis I think it is best if we add a check in a single method, and we add the exception message to logger. I'll be updating your PR within the next minutes to make sure it is included in the next release. Thanks for your PR!

@fl4via I have changed it accordingly. Thanks

msfm:
It's just nit-picking..., but the indentations don't seem to be aligned.

servlet/src/main/java/io/undertow/servlet/UndertowServletMessages.java
```diff
@@ -232,4 +232,7 @@
 
     @Message(id = 10062, value = "No SecurityContext available")
     ServletException noSecurityContextAvailable();
+
+    @Message(id = 10063, value = "This Context has been already destroyed")
+    IllegalStateException nullDeploymentInfo();
```
fl4via: This is the wrong class to add the log message.
I know that having two classes for logs might be confusing, but this class contains errors messages that explain to user that something is missing or wrong on the config or code.
For this type of error, that is internal to the server (i.e., not caused directly by a problem in the deployment or deployment config), we will should use UndertowServletLogger instead.

servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
```diff
@@ -207,11 +207,12 @@ public Void call(HttpServerExchange exchange, ThreadSetupHandler.Action<Void, Ob
 
     @Override
     public String getContextPath() {
-        String contextPath = deploymentInfo.getContextPath();
-        if (contextPath.equals("/")) {
-            return "";
-        }
-        return contextPath;
+            checkDeploymentInfo();
```
fl4via: This type of approach does not work to prevent the NPE. The reason is that we risk having the deploymentInfo set to ull between the execution of checkDeploymentInfo and the next line.

fl4via (commented):
I'm adding these observations to the PR as part of my review but, given that I need to release Undertow within the next hours, I took the freedom to update the PR myself this time.

PR #1025, opened 2021-01-24
[UNDERTOW-1833] Unneccessary code removed from the test.

jstourac:
https://issues.redhat.com/browse/UNDERTOW-1833

This is a trivial change of removal unused and unnecessary code in
the Http2EndExchangeTestCase.java test case.

Apart from that, I slightly changed the way how test fails in case that
HTTP2 connection is not correctly established - I unified this failure
with other failures in the test.

There is no other problem in the test as it is passing just fine.
I just got around this when I was investigating something else.


jstourac:
Regarding to the CI job results - I don't think that failures there are related to changes in this PR.

fl4via:
@jstourac Thanks for your PR! We've been working sistematically between one task and another to fix the failures, they are probably unrelated, I'll double check.

PR #1031, opened 2021-02-01
Added Mimetype for mp4, webm, flac, weba , csv and webp

dextro67:


dextro67:
@stuartwdouglas @fl4via Could you review these changes and comment. My goal here was to add support for mp4 mimetype mostly.

fl4via:
I am for now adding the new feature api change, as I don't think it makes sense backporting support of new mime types. However, if there is need / request to backport it ,we will do so.

PR #1038, opened 2021-02-10
UNDERTOW-1844: Fix getHttpServletMapping when both prefix and extension match

carterkozak:
https://issues.redhat.com/browse/UNDERTOW-1844

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatchesData.java
+++ b/servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatchesData.java
@@ -79,19 +79,15 @@ class ServletPathMatchesData {
             }
         }
         //this should never happen
-        //as the default servlet is aways registered under /*
+        //as the default servlet is always registered under /*
         throw UndertowMessages.MESSAGES.servletPathMatchFailed();
     }
 
     private ServletPathMatch handleMatch(final String path, final PathMatch match, final int extensionPos) {
-        if (match.extensionMatches.isEmpty()) {
+        if (extensionPos == -1 || match.extensionMatches.isEmpty()) {
             return new ServletPathMatch(match.defaultHandler, path, match.requireWelcomeFileMatch);
         }
-        if (extensionPos == -1) {
-            return new ServletPathMatch(match.defaultHandler, path, match.requireWelcomeFileMatch);
-        }
-        final String ext;
-        ext = path.substring(extensionPos + 1, path.length());
+        final String ext = path.substring(extensionPos + 1);
         ServletChain handler = match.extensionMatches.get(ext);
         if (handler != null) {
             return new ServletPathMatch(handler, path, handler.getManagedServlet().getServletInfo().isRequireWelcomeFileMapping());
--- a/servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
@@ -250,15 +250,18 @@ public final class HttpServletRequestImpl implements HttpServletRequest {
                 break;
             case PATH:
                 matchValue = match.getRemaining();
-                if(matchValue.startsWith("/")) {
+                if (matchValue == null) {
+                    matchValue = "";
+                } else if (matchValue.startsWith("/")) {
                     matchValue = matchValue.substring(1);
                 }
                 break;
             case EXTENSION:
-                matchValue = match.getMatched().substring(0, match.getMatched().length() - match.getMatchString().length() + 1);
-                if(matchValue.startsWith("/")) {
-                    matchValue = matchValue.substring(1);
-                }
+                String matched = match.getMatched();
+                String matchString = match.getMatchString();
+                int startIndex = matched.startsWith("/") ? 1 : 0;
+                int endIndex = matched.length() - matchString.length() + 1;
+                matchValue = matched.substring(startIndex, endIndex);
                 break;
             default:
                 matchValue = match.getRemaining();
--- a/servlet/src/test/java/io/undertow/servlet/test/path/MultipleMatchingMappingTestCase.java
+++ b/servlet/src/test/java/io/undertow/servlet/test/path/MultipleMatchingMappingTestCase.java
@@ -1,0 +1,94 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package io.undertow.servlet.test.path;
+
+import io.undertow.servlet.api.ServletInfo;
+import io.undertow.servlet.test.util.DeploymentUtils;
+import io.undertow.testutils.DefaultServer;
+import io.undertow.testutils.HttpClientUtils;
+import io.undertow.testutils.TestHttpClient;
+import io.undertow.util.StatusCodes;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.junit.Assert;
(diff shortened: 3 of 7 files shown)


servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatchesData.java
```diff
@@ -84,14 +84,10 @@ public ServletPathMatch getServletHandlerByPath(final String path) {
...
-        if (match.extensionMatches.isEmpty()) {
+        if (extensionPos == -1 || match.extensionMatches.isEmpty()) {
             return new ServletPathMatch(match.defaultHandler, path, match.requireWelcomeFileMatch);
         }
-        if (extensionPos == -1) {
-            return new ServletPathMatch(match.defaultHandler, path, match.requireWelcomeFileMatch);
-        }
-        final String ext;
-        ext = path.substring(extensionPos + 1, path.length());
+        final String ext = path.substring(extensionPos + 1);
```
carterkozak, line 90: No functional change here, I removed the unnecessary second argument to substring, and combined the two conditionals with the same result

servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
```diff
@@ -255,10 +255,12 @@ public HttpServletMapping getHttpServletMapping() {
                 }
                 break;
             case EXTENSION:
-                matchValue = match.getMatched().substring(0, match.getMatched().length() - match.getMatchString().length() + 1);
-                if(matchValue.startsWith("/")) {
-                    matchValue = matchValue.substring(1);
-                }
+                String remaining = match.getRemaining();
+                String matched = remaining != null ? remaining : match.getMatched();
```
carterkozak, line 259: This line fixes our bug.

Previously `match.getMatched()` returned the matched prefix, even in the case of an extension match, if both types of matches occurred. Now we use remaining if present, falling back to matched if remaining is null.
If we have a remaining segment, that's where we can find the matched extension. If there's no remaining segment, the previous implementation was correct.

servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
```diff
@@ -255,10 +255,12 @@ public HttpServletMapping getHttpServletMapping() {
...
+                String matchString = match.getMatchString();
+                int startIndex = matched.startsWith("/") ? 1 : 0;
+                int endIndex = matched.length() - matchString.length() + 1;
+                matchValue = matched.substring(startIndex, endIndex);
```
carterkozak, line 263: I combined the two subsequent substring operations to avoid re-allocating the matchValue string.

servlet/src/test/java/io/undertow/servlet/test/path/MultipleMatchingMappingTestCase.java
```diff
@@ -0,0 +1,86 @@
...
+                    "Match value:foo\n" +
+                    "Pattern:*.ext", response);
+
+            get = new HttpGet(DefaultServer.getDefaultServerURL() + "/servletContext/foo.ext");
+            result = client.execute(get);
+            Assert.assertEquals(StatusCodes.OK, result.getStatusLine().getStatusCode());
+            response = HttpClientUtils.readResponse(result);
+            Assert.assertEquals("Mapping match:EXTENSION\n" +
+                    "Match value:foo\n" +
+                    "Pattern:*.ext", response);
```
carterkozak, line 72: This case reproduces the StringIndexOutOfBounds failure from the ticket.

servlet/src/test/java/io/undertow/servlet/test/path/MultipleMatchingMappingTestCase.java
```diff
@@ -0,0 +1,86 @@
...
+    public void testGetMappingMatchingExtensionAndPrefix() throws IOException {
+        TestHttpClient client = new TestHttpClient();
+        try {
+            HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/servletContext/path/foo.ext");
+            HttpResponse result = client.execute(get);
+            Assert.assertEquals(StatusCodes.OK, result.getStatusLine().getStatusCode());
+            String response = HttpClientUtils.readResponse(result);
+            Assert.assertEquals("Mapping match:EXTENSION\n" +
+                    "Match value:foo\n" +
+                    "Pattern:*.ext", response);
```
carterkozak, line 64: Without the bugfix, the match value in this case is empty string, however that's because `path` and `.ext` are the same length. if I used a longer prefix, for example replacing `path` with `abcdef` the result would be match value would report `ab` based on a match portion `/abcdef` with the length of `.ext` stripped from the end.

carterkozak: I’m not confident this is correct since it takes into account both prefix and extension matches.
I will verify the specification when I have a moment.

carterkozak: Alright, so the [Servlet 4.0 specification](https://javaee.github.io/servlet-spec/downloads/servlet-4.0/servlet-4_0_FINAL.pdf) section 12.1 states:

> 1. The container will try to find an exact match of the path of the request to the path
of the servlet. A successful match selects the servlet.
> 2. The container will recursively try to match the longest path-prefix. This is done
by stepping down the path tree a directory at a time, using the ’/’ character as a
path separator. The longest match determines the servlet selected.
> 3. If the last segment in the URL path contains an extension (e.g. .jsp), the servlet
container will try to match a servlet that handles requests for the extension. An
extension is defined as the part of the last segment after the last ’.’ character.
> 4. If neither of the previous three rules result in a servlet match, the container will
attempt to serve content appropriate for the resource requested. If a "default"
servlet is defined for the application, it will be used. Many containers provide an
implicit default servlet for serving content.

Based on that, we're erroneously classifying this as an EXTENSION match. This is confirmed by the section 12.2.2 examples, second example.

When I have a moment, I'll reproduce the example cases in this test. It's unclear if we'd want to change the behavior beyond fixing the exception in a patch version release because there's some potential to break existing applications, but I defer to your judgement.

servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
```diff
@@ -250,15 +250,18 @@ public HttpServletMapping getHttpServletMapping() {
                 break;
             case PATH:
                 matchValue = match.getRemaining();
-                if(matchValue.startsWith("/")) {
+                if (matchValue == null) {
+                    matchValue = "";
```
carterkozak, line 254: This fixes an NPE caught by the servlet-spec-examples test case

servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatchesData.java
```diff
@@ -79,22 +80,22 @@ public ServletPathMatch getServletHandlerByPath(final String path) {
...
-            return new ServletPathMatch(match.defaultHandler, path, match.requireWelcomeFileMatch);
-        }
-        final String ext;
-        ext = path.substring(extensionPos + 1, path.length());
-        ServletChain handler = match.extensionMatches.get(ext);
-        if (handler != null) {
-            return new ServletPathMatch(handler, path, handler.getManagedServlet().getServletInfo().isRequireWelcomeFileMapping());
+        MappingMatch matchType = match.defaultHandler.getMappingMatch();
+        // Extension matches take precedence over the default servlet, but not path or exact matches
+        if (matchType != MappingMatch.PATH && matchType != MappingMatch.EXACT) {
```
carterkozak, line 93: Here we wrap the existing code such that EXTENSION mapping doesn't occur when we have any PATH or EXACT match, based on the servlet specification.

carterkozak: Ahh this fails due to filters, which can match both. I think I'm looking at the wrong layer. ServletPathMatches.java looks like the correct place to resolve this issue.

servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
carterkozak: Updated, this comment referred to a change that avoided throwing exceptions, but did not fix the underlying issue diverging from the servlet spec.

servlet/src/test/java/io/undertow/servlet/test/path/MultipleMatchingMappingTestCase.java
carterkozak: I've added the examples from the servlet spec as a new test case, and updated extension mapping to only take precedence over the default servlet, not other path or prefix matches.

fl4via (approved):

fl4via (approved):

PR #1041, opened 2021-02-16
UNDERTOW-1794 DefaultAccessLogReceiver violates Closeable contract

baranowb:
ISSUE: https://issues.redhat.com/browse/UNDERTOW-1794


core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -153,75 +154,60 @@ public void logMessage(final String message) {
...
-        } finally {
-            // change this state to final state
-            stateUpdater.set(this, 3);
-            //check to see if there is still more messages
-            //if so then run this again
-            if (!pendingMessages.isEmpty() || forceLogRotation) {
-                if (stateUpdater.compareAndSet(this, 3, 1)) {
-                    logWriteExecutor.execute(this);
+            for (int i = 0; i < 1000 && !this.closed && !pendingMessages.isEmpty(); ++i) {
+                final String msg = pendingMessages.poll();
```
baranowb, line 184: It might be better to use peek()+remove() here.

fl4via: @baranowb because this class has concurrent access, poll is better than peek() + remove(). Between the moment we invoke peek() and the moment we invoke remove, we risk having another thread executing flushAndTerminate. This is why we need to run poll() here.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -309,11 +325,59 @@ public void rotate() {
...
-        if (stateUpdater.compareAndSet(this, 0, 1)) {
-            logWriteExecutor.execute(this);
+        this.closed = true;
+        if (this.stateUpdater.compareAndSet(this, 0, 3)) {
+            flushAndTerminate();
+            return;
+        } else {
+            // state[1,2] - scheduled or running, attempt schedule hijack
+            if (this.stateUpdater.compareAndSet(this, 1, 3)) {
+                //this means this thread raced against scheduled run(). run() will exist ASAP
```
fl4via, line 342: run() will exit

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -309,11 +325,59 @@ public void rotate() {
         }
     }
 
+    @SuppressWarnings("static-access")
     @Override
     public void close() throws IOException {
-        closed = true;
-        if (stateUpdater.compareAndSet(this, 0, 1)) {
-            logWriteExecutor.execute(this);
+        this.closed = true;
```
fl4via, line 335: method must take into account that the state could change between one if and the other. So, there is a possibility that all stateUpdater checks return false.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -309,11 +325,59 @@ public void rotate() {
...
+            // either failed race to 1->3 or we were in 2. We have to wait here sometime.
+            // wait ~1s, if situation does not clear up, try dumping stuff
+            for(int i=0; i<20;i++) {
+                try {
+                    Thread.currentThread().sleep(50);
+                } catch (InterruptedException e) {
+                    // TODO Auto-generated catch block
+                    e.printStackTrace();
+                }
+                if(this.stateUpdater.get(this) == 3) {
```
fl4via, line 356: there is a possibility that we exit run without having the status 3. The reason is that run does not work as a single unit, we are not synchronized. So, if we set closed to true too late in the game, we risk being at 2 while sleeping and waking up in a 0 or 1 scenario

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -309,11 +325,59 @@ public void rotate() {
...
+                    break;
+                }
+            }
+            this.stateUpdater.set(this, 3);
```
fl4via: we can never invoke a set without a compareAndSet, because we don't know what the state value is at this point

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -309,11 +325,59 @@ public void rotate() {
...
+            flushAndTerminate();
+        }
+    }
+
+    protected void flushAndTerminate() {
+        try {
+            while (!this.pendingMessages.isEmpty()) {
+                final String msg = this.pendingMessages.poll();
+                // TODO: clarify this, how is this possible?
```
fl4via, line 372: It might not be possible. Lets wait for the final version of the code of run() and close() to decide this

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -309,11 +325,59 @@ public void rotate() {
...
+            return;
+        } else {
+            // state[1,2] - scheduled or running, attempt schedule hijack
+            if (this.stateUpdater.compareAndSet(this, 1, 3)) {
+                //this means this thread raced against scheduled run(). run() will exist ASAP
+                //as 1->2 wont be possible, we are at 3 and this.closed
+                flushAndTerminate();
+                return;
+            }
+            // either failed race to 1->3 or we were in 2. We have to wait here sometime.
```
fl4via, line 347: this part needs more thought. Invoking sleep without a check for 2 risks us being in another state. And we need to make this method idempotent, i.e., if it is being closed from 2 different threads, only one of them should run.

fl4via (changes requested):

fl4via:
as this is a new feature, merging of this PR will require synchronization with the RFE process. I'll assist in having all the process followed as fast as we can to have this merged the soonest.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
baranowb: There is trade off. It is possible that XNIO thread will be terminated between poll() and write(), leaking one message.
If its peek()+remove(), it might double entry, but not loose it. Its a trade off.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
baranowb: yes, as it did not reach any functional block, just might be sleeping, so as soon as it will kick in( either from dormancy or concurrency induced async) it will not enter any IO block.

fl4via: I see

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: You're right about it. My question is now is hot to prevent both scenarios. Is there some way we can synchronize just a little bit to prevent a duplicate entry in the log? I think we must prevent an output that will be considered buggy by users if someone finds a duplicate entry after shutting down the server.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -153,75 +154,66 @@ public void logMessage(final String message) {
      */
     @Override
     public void run() {
+        //check if we can transition to 2. If so, perform tasks in small chunks and check this.closed.
+        //move into 3 if(this.closed) and terminate run()
         if (!stateUpdater.compareAndSet(this, 1, 2)) {
             return;
         }
-        if (forceLogRotation) {
-            doRotate();
```
fl4via, line 160: I've been thinking about this, and if you remove this, you're breaking the rotate() method contract

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -243,43 +235,73 @@ void awaitWrittenForTest() throws InterruptedException {
         }
     }
 
-    private void writeMessage(final List<String> messages) {
-        if (System.currentTimeMillis() > changeOverPoint) {
-            doRotate();
+    private void writeMessage(final String message) {
+        //NOTE: is there a need to rotate on write?
+        //if (System.currentTimeMillis() > changeOverPoint) {
```
fl4via, line 244: Yes, but only once per run. Now that this method is called per message it is no longer the adequate place for that.

fl4via: Probably we should add the limit to 1000 here? I think it would perform a bit better.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: resolved.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: this still needs to be updated

fl4via (changes requested):

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: My bad, there is a check right below, so the rotation algorithm is working correctly in the PR.

core/src/main/java/io/undertow/UndertowLogger.java
```diff
@@ -452,4 +453,21 @@ void nodeConfigCreated(URI connectionURI, String balancer, String domain, String
     @LogMessage(level = DEBUG)
     @Message(id = 5098, value = "GSSAPI negotiation failed for %s")
     void failedToNegotiateAtGSSAPI(final HttpServerExchange exchange, final @Cause Throwable e);
+
+    @LogMessage(level = DEBUG)
+    @Message(id = 5099, value = "Access Log Worker failed to transition gracefuly.")
+    void accessLogWorkerFailureOnTransition();
+
+    @LogMessage(level = DEBUG)
+    @Message(id = 6000, value = "Access Log Worker failed to reschedule.")
```
baranowb, line 462: Big brain math....

baranowb: On the other hand - no one noticed :)

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -153,75 +154,66 @@ public void logMessage(final String message) {
...
-            // change this state to final state
-            stateUpdater.set(this, 3);
-            //check to see if there is still more messages
-            //if so then run this again
-            if (!pendingMessages.isEmpty() || forceLogRotation) {
-                if (stateUpdater.compareAndSet(this, 3, 1)) {
-                    logWriteExecutor.execute(this);
+            for (int i = 0; i < 1000 && !this.closed && !pendingMessages.isEmpty(); ++i) {
+                final String msg = pendingMessages.peek();
+                if (msg == null) {
```
fl4via, line 185: it is not supposed to be null, because you are checking it is not empty

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -153,75 +154,66 @@ public void logMessage(final String message) {
...
-        } finally {
-            // change this state to final state
-            stateUpdater.set(this, 3);
-            //check to see if there is still more messages
-            //if so then run this again
-            if (!pendingMessages.isEmpty() || forceLogRotation) {
-                if (stateUpdater.compareAndSet(this, 3, 1)) {
-                    logWriteExecutor.execute(this);
+            for (int i = 0; i < 1000 && !this.closed && !pendingMessages.isEmpty(); ++i) {
+                final String msg = pendingMessages.peek();
```
fl4via, line 184: I might be looking too much into details here, but why peeking and removing instead of just polling?

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -153,75 +154,66 @@ public void logMessage(final String message) {
...
+                writeMessage(msg);
+
+                // NOTE:this is very similar to remove(), but without screenNull
+                // at best, it will work like poll/remove, at worst, will do nothing
+                if (!pendingMessages.remove(msg)) {
+                    break;
+                }
+            }
+        }finally {
+            // flush what we might have
```
fl4via, line 197: this is a very good point

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: run() will exit -> just a typo to be fixed in the comment

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -153,75 +154,66 @@ public void logMessage(final String message) {
...
-                    } finally {
-                        // reset the state to 0 again finally
-                        stateUpdater.set(this, 0);
+            if(this.closed) {
+                if (!stateUpdater.compareAndSet(this, 2, 3)) {
+                    UndertowLogger.ROOT_LOGGER.accessLogWorkerFailureOnTransition();
+                }
+                return;
+            } else {
+                if (!pendingMessages.isEmpty() || forceLogRotation) {
```
fl4via, line 212: I have been thinking... we still need a state for being at the finally block, as it prevents us from having more than one thread scheduling a new execution of run at the same time

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -59,7 +58,7 @@ public class DefaultAccessLogReceiver implements AccessLogReceiver, Runnable, Cl
     //0 = not running
     //1 = queued
     //2 = running
-    //3 = final state of running (inside finally of run())
```
fl4via, line 62: I added a comment down below that we will still need state 3 to be used to keep tracking of being at finally of run)
That being said what are we trying to achieve with the new state 3? (It will have to be moved to 4 if we decide to keep it)> Are we trying to prevent run from doing anything after closed is invoked? Is that its purpose?

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -139,12 +138,14 @@ private void calculateChangeOverPoint() {
 
     @Override
     public void logMessage(final String message) {
+        if(closed) {
+            //Log handler is closing, other resources should as well, there shouldn't
+            //be resources served that required this to log stuff into AL file.
+            throw UndertowMessages.MESSAGES.failedToLogAccessOnClose();
+        }
         this.pendingMessages.add(message);
-        int state = stateUpdater.get(this);
```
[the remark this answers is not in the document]
fl4via: btw, we can safely read state directly from the state field here, and that's the cheapest option here instead of reading get into a private field

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: I still stand my ground, I think the granularity of checking 1000 here performs better.

fl4via (changes requested):
There are some pending issues here to consider.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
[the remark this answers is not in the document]
fl4via: Unfortunately github picked an old coment that I thought I deleted. So, here goes my final conclusion on this: 

```
if(myIntegerField == whateverValueINeedToTest) {
    if (!myIntegerFieldUpdater.compareAndSet(whateverValueINeedToTest, newValue) {
        return;
    }
}
``` 

The one above is my preferred approach for this kind of checks. For one thing, myIntegerField being volatile is not subject to word tearing and we can safely refer to that instead of invoking myIntegerFieldUpdater.get(), so that's one less invocation. Also, when dealing with that, if the field's value is not the expected one, you exit (the exit part is not shown in the above snippet) without having to invoke the field updater, so it is cheaper.

All that being said, I am in favour of replacing it by just an invocation to compareAndSet like you did in this PR, because, in this particular run() method, for the expected use of DefaultAccessLogReceiver, the value is expected to be whateverValueINeedToTest in 99% of the times it reaches this code, so you probably were going to have to pay the price of invoking compareAndSet anyway.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: okay, it is clearly stated in the Jira: we want to make sure that close returns only after it is fully closed. The question that remains is if we need a new state for it, I think we need so we can fully skip the rest of the execution of run() in case it is closing, without having to resort to verifying the value of close.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -139,12 +138,14 @@ private void calculateChangeOverPoint() {
 
     @Override
     public void logMessage(final String message) {
+        if(closed) {
+            //Log handler is closing, other resources should as well, there shouldn't
+            //be resources served that required this to log stuff into AL file.
+            throw UndertowMessages.MESSAGES.failedToLogAccessOnClose();
+        }
```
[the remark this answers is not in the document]
baranowb: Well. Its a tricky scenario. When this happens, every connection/exchange should/will be purged. 
When exception is thrown, it potentially wont allow to close connection.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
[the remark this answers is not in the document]
baranowb: In this place its a hail marry sort of situation.  If access log is running, run(), will pick up excess messages and schedule a run, if not, this will do it. So it will either be irrelevant and needs to synced up just in case, or will schedule new run.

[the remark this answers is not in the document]
fl4via: Agreed. I'm marking this as resolved.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
baranowb: in case there is some intermitent IO failure, it should not purge entries without writing them first. Though, in such case, some overflow mechanism should be in place as well?

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
baranowb: how? if there is nothing in pendingMessages, it does not need scheduling, if someone writes, it will be able to schedule. In reality, pendingMessages are filled before state change, so its just a race between logMessage thread and run() to schedule new execution.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
baranowb: Also "i < 1000 && !this.closed &&" checking this in loop might not be best, should be enough to loop through and check at the end IMHO

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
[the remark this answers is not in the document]
baranowb: Also, note that its expected behavior from closed resource.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
baranowb: I dont think so. Its a bit different FSM/CAS that was present before. It relied only on worker threads to dump messages even when handler has been closed and once its done close. This cant be done now since those worker threads may be disabled( upstream issue)

PrarthonaPaul (commented):
Hi @baranowb 
Thanks! The PR looks good, I have added some minor comments.

core/src/main/java/io/undertow/UndertowLogger.java
```diff
@@ -468,4 +469,21 @@ void nodeConfigCreated(URI connectionURI, String balancer, String domain, String
...
+    @LogMessage(level = DEBUG)
+    @Message(id = 5104, value = "Access Log Worker failed to reschedule.")
+    void accessLogWorkerFailureOnReschedule();
+
+    @LogMessage(level = DEBUG)
+    @Message(id = 5105, value = "Access Log Worker did not terminate cleanly.")
+    void accessLogWorkerNoTermination();
+
+    @LogMessage(level = DEBUG)
+    @Message(id = 5106, value = "Interruption in close()")
```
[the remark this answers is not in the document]
baranowb: hmm, weirdly it there.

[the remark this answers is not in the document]
baranowb: ok, this confused me a bit. seems like in rush of fixing/updating RFEs I did not push this one.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
baranowb: Not unless we control threat termination, which we dont.

ropalka (approved):

ropalka (changes requested):
I approved it by mistake, I oversaw the second commit. Leaving as changes requested until all Flavia's concerns are sorted out.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: I don't know what I meant by limit of 1000. I think it looks good the way it is because we are checking for rotation in run already, and, as I said, checking per message is not the ideal place for this.

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
fl4via: I think it is very unlikely, but still somethiing to safeguard against. Because we are on a best effort basis in graceful shutdown when closing, we might leave the loop in line 372 still in state 2. If that occurs, we have a race when trying to write the last messages. So, it could happens that poll returns null (and it could also result in a duplicate write of the same message).

fl4via (changes requested):
Please rebase the PR against main and I think you can delete the question on how is poll returning null possible. The other notes I think we can leave for future reference, unless you are certain the questions have been fully answered, in that case, the cleaner the better. Thanks for looking into this!

PR #1048, opened 2021-03-02
[UNDERTOW-1855] Prevent multiple service loading of ExchangeAttribute…

fl4via:
…Builder and PredicateHandlerBuilder services for a same classloader

Jira: https://issues.redhat.com/browse/UNDERTOW-1855


fl4via:
I'll be reviewing whether this could be put in WF instead of here, and I'll do some more research on WF side for other services being loaded as well.

PR #1062, opened 2021-03-23
[UNDERTOW-1868] Use proper stream Id in HttpFramePriority to allow mo…

baranowb:
…re than initial push
Issue: https://issues.redhat.com/browse/UNDERTOW-1832

### DIFF
--- a/core/src/main/java/io/undertow/protocols/http2/Http2PushPromiseStreamSinkChannel.java
+++ b/core/src/main/java/io/undertow/protocols/http2/Http2PushPromiseStreamSinkChannel.java
@@ -53,4 +53,8 @@ public class Http2PushPromiseStreamSinkChannel extends Http2DataStreamSinkChanne
         return bytes;
     }
 
+    public int getPushedStreamId() {
+       return pushedStreamId;
+    }
+
 }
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/protocols/http2/Http2PushPromiseStreamSinkChannel.java
```diff
@@ -53,4 +53,8 @@ protected int grabFlowControlBytes(int bytes) {
         return bytes;
     }
 
+    public int getPushedStreamId() {
```
ropalka, line 56: Please do no introduce new getPushedStreamId() method but override existing getStreamId() one instead

ropalka (changes requested):

core/src/main/java/io/undertow/protocols/http2/Http2PushPromiseStreamSinkChannel.java
baranowb: 1. Im a away from my station, but Im fairly sure this method is being used to form header of a frame or something that is crucial
2. Use of this getter name would void its contract. it would no longer return actual stream ID which will convey frame, in this case it would return some future ID that might not manigest

As is, Id vote to introduce new getter, as it does not get in way of existing code and makes it clear what is being vetted.

baranowb: Though, just to be thorough, I will check if it blows up in the morning.

baranowb: Yep, just as I thought: ERR_HTTP2_PROTOCOL_ERROR  booom.

Examples: 
getChannel().removeStreamSink(getStreamId());

or

Http2ProtocolUtils.putInt(firstBuffer, getStreamId());  from Http2DataStreamSinkChannel#createFrameHeaderImpl

ropalka:
Thanks for clarification @baranowb

ropalka (approved):

PR #1079, opened 2021-04-01
[UNDERTOW-1876] security-manager and reflection permissions in Direct…

soul2zimate:
…ByteBufferDeallocator/undertow

Fix made in https://issues.redhat.com/browse/UNDERTOW-1558 only covers JDK version < 9, this completes it for JDK >= 9 case.

Issue: https://issues.redhat.com/browse/UNDERTOW-1876
2.1.x PR: https://github.com/undertow-io/undertow/pull/1080
2.0.x PR: https://github.com/undertow-io/undertow/pull/1081

### DIFF
--- a/core/src/main/java/io/undertow/server/DirectByteBufferDeallocator.java
+++ b/core/src/main/java/io/undertow/server/DirectByteBufferDeallocator.java
@@ -43,8 +43,7 @@ public final class DirectByteBufferDeallocator {
         } else {
             try {
                 tmpUnsafe = getUnsafe();
-                tmpCleanerClean = tmpUnsafe.getClass().getDeclaredMethod("invokeCleaner", ByteBuffer.class);
-                tmpCleanerClean.setAccessible(true);
+                tmpCleanerClean = getDeclaredMethod(tmpUnsafe, "invokeCleaner", ByteBuffer.class);
                 supported = true;
             } catch (Throwable t) {
                 UndertowLogger.ROOT_LOGGER.directBufferDeallocatorInitializationFailed(t);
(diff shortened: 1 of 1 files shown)


soul2zimate:
Are any CI jobs running with security manager enabled ?  Because the change only make a difference w/ and w/o security manager enabled.

Also, This change only made for JDK version >= 9, it should not affect any Undertow CI / JDK 8 jobs.

fl4via (approved):

boris-unckel:
Hello @fl4via @soul2zimate 
this change seems strange to me for security. In [https://issues.redhat.com/browse/UNDERTOW-1558](https://issues.redhat.com/browse/UNDERTOW-1558) is an attachment to reproduce a security manager issue. A META-INF/permissions.xml is missing. Obviously the security manager will throw an exception for restricted methods - nothing is allowed.
Maybe I miss something in the case.
Kind regards
Boris
fyi @bstansberry

soul2zimate:
The `permissions.xml` allows you to define custom security checks for the app/deployment.

In regards to `getDeclaredMethod()` and  `setAccessible` in this example, they require the caller domain to have proper "core"  Java permissions for successful execution. They needs to be wrapped in a `AccessController.doPrivileged()` to mark those calls as "privileged" so that such insufficient permissions from the app/deployment are ignored.

boris-unckel:
@soul2zimate 
> ... to mark those calls as "privileged" so that such insufficient permissions from the app/deployment are ignored.
Thanks for your answer. I'm still not convinced. If it is possible to grant permissions, why should you build the system in a way that makes a grant not necessary? The usecase of activating the Java Security Manager is to enable checks. Someone who expects "activated checks without need to grant rights" is wrong and can simply not enable the JSM.

soul2zimate:
In the example, only the calling of `getDeclaredMethod()` and `setAccessible()` is executed with all the permissions. User can still activate security manager to check other customized permissions if wanted.

fl4via:
@boris-unckel From my pov, anything that is internal and is required to run a regular application should run without throwing unexpected exceptions. Hence, if that internal code runs an action that requires special permission, it should be run from inside a privileged block, unless a malicious attacker can take advantage of that (which is not the case here). Without this fix, a simple attempt to read the input stream in the servlet can cause a security exception. It occurs if memory has to be freed in the background during the read attempt. 
The final user is not supposed to know this specific detail, specially it being something that occurs occasionally. Nor is he expected to know that he has to grant permission to that specific code. This is why this fix is being proposed.

PR #1085, opened 2021-04-06
[UNDERTOW-1877] HTTP2 implementation returns PUSH PROMISES for frames…

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-1877

Fix to not send push promises if the initiator is a previous push promise (only odd number frames, client requests, are allowed to push). The push builder is not returned (null) in this condition. This is the modification in `Http2ClientConnection.java`.

Other changes were needed to manage push promises at client side and do the test. A client can receive a request response (odd number) or a push promise if enabled (even). A server can only receive odd IDs (client requests). So the `lastGoodStreamId` should be differentiated, a new `lastAssignedStreamOtherSide` is used for the task. The pushed streams are also added to current streams in the client with a new method `addPushPromiseStream`. Test added with these changes.


rmartinc:
Added some more synchs to make spotbugs happy. ;-)

fl4via (approved):

PR #1103, opened 2021-04-18
[UNDERTOW-1834] Fix for IBM JDK and HTTP2/ALPN.

jstourac:
https://issues.redhat.com/browse/UNDERTOW-1834

There is a difference in IBM JDK TLS ciphersuite names versus their
Oracle JDK counterparts, see e.g. [1].

This discrepancy causes issue in usage of HTTP2 with IBM JDK since
we check enabled ciphersuite based on the name in Oracle JDK variant
only. So if there is an incomming connection from client that offers
IBM names of the ciphersuite, we simply fail to use ALPN and HTTP2
with it consequently.

This change should fix such issue.

[1] https://www.ibm.com/support/knowledgecenter/SSFKSJ_9.0.0/com.ibm.mq.dev.doc/q113210_.htm


ropalka (approved):

ropalka:
Thanks @jstourac !

PR #1110, opened 2021-04-21
[UNDERTOW-1870] - make AsyncContext.timeout configurable per deployment

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1870

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
+++ b/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
@@ -86,6 +86,7 @@ public class DeploymentInfo implements Cloneable {
     private ConfidentialPortManager confidentialPortManager;
     private boolean allowNonStandardWrappers = false;
     private int defaultSessionTimeout = 60 * 30;
+    private long defaultAsyncContextTimeout = 30000;
     private ConcurrentMap<String, Object> servletContextAttributeBackingMap;
     private ServletSessionConfig servletSessionConfig;
     private String hostName = "localhost";
@@ -316,6 +317,18 @@ public class DeploymentInfo implements Cloneable {
      */
     public DeploymentInfo setDefaultSessionTimeout(final int defaultSessionTimeout) {
         this.defaultSessionTimeout = defaultSessionTimeout;
+        return this;
+    }
+
+    public long getDefaultAsyncConextTimeout() {
+        return defaultAsyncContextTimeout;
+    }
+
+    /**
+     * @param defaultAsyncContextTimeout The default async context timeout, in milliseconds
+     */
+    public DeploymentInfo setDefaultAsyncConextTimeout(final long defaultAsyncContextTimeout) {
+        this.defaultAsyncContextTimeout = defaultAsyncContextTimeout;
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/UndertowOptions.java
```diff
@@ -77,6 +77,18 @@
...
+     * Default value( miliseconds ) for async context timeout.
+     * @see {@link javax.servlet.AsyncContext#setTimeout}
+     */
+    public static final Option<Long> ASYNC_CONTEXT_TIMEOUT = Option.simple(UndertowOptions.class, "ASYNC_CONTEXT_TIMEOUT", Long.class);
+
+    /**
+     * Default value of {@link #ASYNC_CONTEXT_TIMEOUT}
+     * @see {@link javax.servlet.AsyncContext#setTimeout}
+     */
+    public static final long DEFAULT_ASYNC_CONTEXT_TIMEOUT = 30000;
```
msfm, line 90: I think `ASYNC_CONTEXT_TIMEOUT` and `DEFAULT_ASYNC_CONTEXT_TIMEOUT` are not needed as UndertowOptions. If I understand correctly, `UndertowOptions` is basically used for specifying the listener-level options.
I think just adding an API to `DeploymentInfo` is enough for the deployment-level option because WildFly (JBoss EAP) can invoke the `DeploymentInfo` API from `UndertowDeploymentInfoService` to configure `<servlet-container>` setting as the default deployment-level options.

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
```diff
@@ -96,6 +97,7 @@
     private ConfidentialPortManager confidentialPortManager;
     private boolean allowNonStandardWrappers = false;
     private int defaultSessionTimeout = 60 * 30;
+    private long defaultAsyncContextTimeout = AsyncContextImpl.DEFAULT_ASYNC_CONTEXT_TIMEOUT;
```
msfm, line 100: Isn't `DeploymentInfo` API better not to depend on a concrete implementation like `AsyncContextImpl` as possible?
I think it's enough to have the default value of `defaultAsyncContextTimeout` (30000) in `DeploymentInfo` like `defaultSessionTimeout` does.

baranowb: if it was separate API package, Id agree. In this case Id consider this ok. For sake of transparency its good to have such values defined with jdoc, so users can look it up, rather search through source.

fl4via: @baranowb I agree with @msfm. One extra point here is keeping consistency. If we were to add final values for timeouts, we would do it in the entire code. If we already have a defaultSessionTimeout defined without a constant field in line 99 above the line you added, it makes the code more difficult to read because it fails to follow a pattern.

baranowb: I kind of dont agree, since some options follow this pattern. Just because others dont makes no sense to ignore it. Though I did update PR as requested.

fl4via (approved):

ropalka (approved):

PR #1113, opened 2021-04-21
[UNDERTOW-1864] EAP returns 403 even after adding the welcome file to unmanaged exploded deploy

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-1864

Summary of changes:

1. There are two caches now, one that contains matches that depends on welcome files and can change and one that is fixed. Trying to maximize performance.
2. The cache that can change uses listeners and TTL (if enabled) to detect changes.
3. The TTL was added to the deployment config. So we need a minor change in wildfly to set the new field in the deployment (no schema change, the value is already there) to really activate this with the component upgrade.
4. Tests added to check welcome files modifications with and without file listeners.

@fl4via check this and tell me how you see it. We can do different things but I like this idea more. We need to backport the fix to 2.0 and 2.1 branches. I'll do when we think it's ready.

PR for 2.1.x: https://github.com/undertow-io/undertow/pull/1145
PR for 2.0.x:  https://github.com/undertow-io/undertow/pull/1144

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
+++ b/servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
@@ -24,6 +24,7 @@ import static io.undertow.servlet.handlers.ServletPathMatch.Type.REWRITE;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -33,10 +34,14 @@ import java.util.Set;
 import javax.servlet.DispatcherType;
 import javax.servlet.http.MappingMatch;
 
+import io.undertow.UndertowLogger;
 import io.undertow.server.HandlerWrapper;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.handlers.cache.LRUCache;
+import io.undertow.server.handlers.resource.CachingResourceManager;
 import io.undertow.server.handlers.resource.Resource;
+import io.undertow.server.handlers.resource.ResourceChangeEvent;
+import io.undertow.server.handlers.resource.ResourceChangeListener;
 import io.undertow.server.handlers.resource.ResourceManager;
 import io.undertow.servlet.UndertowServletMessages;
 import io.undertow.servlet.api.Deployment;
@@ -64,12 +69,48 @@ public class ServletPathMatches {
 
     private volatile ServletPathMatchesData data;
 
-    private final LRUCache<String, ServletPathMatch> pathMatchCache = new LRUCache<>(1000, -1, true); //TODO: configurable
+    // cache for fixed data matches (no welcome files involved)
+    private final LRUCache<String, ServletPathMatch> pathMatchCacheFixed;
+
+    // cache for resource matches that can change
+    private final LRUCache<String, ServletPathMatch> pathMatchCacheResources;
 
     public ServletPathMatches(final Deployment deployment) {
         this.deployment = deployment;
         this.welcomePages = deployment.getDeploymentInfo().getWelcomePages().toArray(new String[deployment.getDeploymentInfo().getWelcomePages().size()]);
         this.resourceManager = deployment.getDeploymentInfo().getResourceManager();
+        this.pathMatchCacheFixed = new LRUCache<>(1000, -1, true);
+        this.pathMatchCacheResources = new LRUCache<>(1000,
+                resourceManager instanceof CachingResourceManager? ((CachingResourceManager) resourceManager).getMaxAge() : -1, true);
+        // add change listener for welcome pages
+        if (this.resourceManager.isResourceChangeListenerSupported()) {
+            try {
+                this.resourceManager.registerResourceChangeListener(new ResourceChangeListener() {
+
+                    @Override
+                    public void handleChanges(Collection<ResourceChangeEvent> changes) {
+                        for (ResourceChangeEvent change : changes) {
+                            // only check added or removed resources that can change welcome files
+                            if (change.getType() != ResourceChangeEvent.Type.MODIFIED) {
+                                String path = "/" + change.getResource();
+                                // remove direct match and directory match
+                                pathMatchCacheResources.remove(path);
+                                pathMatchCacheResources.remove(path + "/");
+                                // check welcome pages
+                                for (String welcomePage: welcomePages) {
+                                    if (path.endsWith("/" + welcomePage)) {
+                                        String pathToUpdate = path.substring(0, path.length() - welcomePage.length());
+                                        pathMatchCacheResources.remove(pathToUpdate);
+                                    }
+                                }
+                            }
+                        }
+                    }
+                });
+            } catch (Exception e) {
+                UndertowLogger.ROOT_LOGGER.couldNotRegisterChangeListener(e);
+            }
+        }
     }
 
     public void initData(){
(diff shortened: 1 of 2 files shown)


rmartinc:
@fl4via Did you take a look to this? WDYT?

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
```diff
@@ -1399,6 +1400,15 @@ public void setPreservePathOnForward(boolean preservePathOnForward) {
         this.preservePathOnForward = preservePathOnForward;
     }
 
+    public int getFileCacheTimeToLive() {
```
fl4via, line 1403: I think we have two different Jiras in this commit: one of them is adding support to configuring the cache parameters, the other one is fixing the bug

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
```diff
@@ -1399,6 +1400,15 @@ public void setPreservePathOnForward(boolean preservePathOnForward) {
...
+        return fileCacheTimeToLive;
+    }
+
+    public DeploymentInfo setFileCacheTimeToLive(int fileCacheTimeToLive) {
```
fl4via, line 1407: Is this here just for the purpose of writing the test case, or is there an intention of making this configurable to the server users?

servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
```diff
@@ -64,12 +68,49 @@
 
     private volatile ServletPathMatchesData data;
 
-    private final LRUCache<String, ServletPathMatch> pathMatchCache = new LRUCache<>(1000, -1, true); //TODO: configurable
+    private static final int CACHE_SIZE = Integer.getInteger("io.undertow.servlet.path.matches.cacheSize", 1000);
+
+    // cache for fixed data matches (no welcome files involved)
+    private final LRUCache<String, ServletPathMatch> pathMatchCacheFixed;
```
fl4via, line 74: Do you think that the performance penalty to be paid at the occasion when someone updates an exploded war is worth the penalty paid in memory, for having the two caches?

servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
```diff
@@ -64,12 +68,49 @@
 
     private volatile ServletPathMatchesData data;
 
-    private final LRUCache<String, ServletPathMatch> pathMatchCache = new LRUCache<>(1000, -1, true); //TODO: configurable
+    private static final int CACHE_SIZE = Integer.getInteger("io.undertow.servlet.path.matches.cacheSize", 1000);
```
fl4via, line 71: Ditto for the two Jiras, here we are adding support to setting cache size, it is not the bug fix.

fl4via (changes requested):
At a minimum we need to split this into two PRs, one with the feature request for setting up the cache config, another one for fixing the bug, I'm not sold on the two caches approach, lets discuss this.

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
rmartinc: @fl4via in the end we need to know the TTL used and apply the same value in the cache inside `ServletPathMatches`. Initially I just tried something like the following:

```java
this.pathMatchCache = new LRUCache<>(1000, this.resourceManager instanceof CachingResourceManager? ((CachingResourceManager) this.resourceManager).getMaxAge() : -1, true);
```

So if it was a caching one I retrieved the TTL from it and if not -1 (cache forever) was used. But I think there are more complicated situations that can use a `DelegatingResourceManager` ([here](https://github.com/wildfly/wildfly/blob/23.0.2.Final/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentInfoService.java#L568-L577)). And that's way I considered better to add the TTL in the deployment info and be sure we are using it correctly.

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
rmartinc: Same explanation than before. It's needed to get the TTL from somewhere because I considered using `CachingResourceManager` was not safe.

servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
rmartinc: In the default configuration I think the penalty is almost 0. Because the default configuration is -1 (cache forever) for deployed apps and 2000 (2s cache) for exploded apps (which are not recommended). So in this default conf we are good, the watch listeners are used (changes are detected in the welcome pages correctly) and everything is cached forever except if they changes.

The only problem IMO is if the customer has modified the `file-cache-time-to-live` in the `servlet-container`. In that case we are always using a TTL (no matter the type of app) and invalidating all the entries after that time. But I think we are always recommending set this to -1 and use the watch service instead.

So, I'm not sure. That's why I decided to go with this solution. But yeah, I can understand that using just one cache is quite safe. I can revert to one cache if you think it's not worthy easily.

servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
rmartinc: Yes, yes, I just added this because it makes sense. This can be removed for this JIRA. :-)

servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
fl4via: @rmartinc I meant memorywise, having the extra cache. Reviewing the code that uses this class, though, I see that's totally neglectable, let's keep the two caches

servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
fl4via: So, as we discussed in the chat, let's split this in two. As soon as the PR is updated I'll merge it, it will go in the next release. Thanks a lot for your PR btw!

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
rmartinc: Finally we are going back to the cast idea to get the TTL from the `CachingResourceManager`. We don't want to change the `DeploymentInfo` API.

servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
rmartinc: OK, removed the CACHE_SIZE from this PR!

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
rmartinc: The changes in teh `DeploymentInfo` have ben removed and teh TTL is obtained from the `resourceManager` if it's a `CachingResourceManager, if not -1 (cache forever) is used as before.

fl4via: Thank you @rmartinc!

fl4via (approved):

fl4via (approved):

PR #1117, opened 2021-04-27
[UNDERTOW-1886] [master] Request dispatcher is returned when the path points to outside the servlet context

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-1886

The `CanonicalPathUtils.canonicalize` now receives an optional boolean to return null when out of context. It's only used in the request dispatcher related methods. A system property `-Dio.undertow.REQUEST_DISPATCHER_NULL_DISALLOWED=true` restores the old behavior just in case. Tests addded.

PR for master.
PR for 2.1.x: https://github.com/undertow-io/undertow/pull/1118
PR for 2.0.x: https://github.com/undertow-io/undertow/pull/1119

### DIFF
--- a/core/src/main/java/io/undertow/util/CanonicalPathUtils.java
+++ b/core/src/main/java/io/undertow/util/CanonicalPathUtils.java
@@ -33,17 +33,21 @@ public class CanonicalPathUtils {
 
 
     public static String canonicalize(final String path) {
+        return canonicalize(path, false);
+    }
+
+    public static String canonicalize(final String path, final boolean nullAllowed) {
         int state = START;
         for (int i = path.length() - 1; i >= 0; --i) {
             final char c = path.charAt(i);
             switch (c) {
                 case '/':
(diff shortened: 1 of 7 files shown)


fl4via (changes requested):
We need to remove the system property, I don't see a reason for being non-spec compliant here and offering that alternative to users

rmartinc:
@fl4via I added the property just in case (it can be removed in that case) but the `CanonicalPathUtils.canonicalize` is used in a lot of places that maybe now don't expect a null returned. That's why I created a second variant with the `nullAllowed` parameter that is used for the request dispatcher. If you are sure that we can return null anywhere please go ahead with your modification.

fl4via:
There is an intermitent failure in DefaultServletCachingListenerTestCase in JDK11 proxy-ajp scenario.
I created this Jira to track it: https://issues.redhat.com/browse/UNDERTOW-1895

PR #1146, opened 2021-05-24
UNDERTOW-1894 - Better handling for leading ? on query string

bdw429s:
Pull for this ticket: https://issues.redhat.com/browse/UNDERTOW-1894
2.0.x PR: #1168

### DIFF
--- a/core/src/main/java/io/undertow/server/HttpServerExchange.java
+++ b/core/src/main/java/io/undertow/server/HttpServerExchange.java
@@ -576,9 +576,9 @@ public final class HttpServerExchange extends AbstractAttachable {
     public HttpServerExchange setQueryString(final String queryString) {
         // Clean leading ?
         if( queryString.length() > 0 && queryString.charAt(0) == '?' ) {
-            this.queryString = queryString.substring(1);
+                this.queryString = queryString.substring(1);
             } else {
-            this.queryString = queryString;
+                this.queryString = queryString;
             }
         return this;
     }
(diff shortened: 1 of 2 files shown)


bdw429s:
@fl4via @stuartwdouglas I see there is a test in the `SetAttributeTestCase` class that is failing, but I believe this test case is **incorrect**.   Instead of the asserting checking for what the response should have been, it simply was written to check for the incorrect behavior.  The `testRewrite()` test asserts that `exchange.getQueryString()` should be return `?bar=a&woz=b` but this is incorrect per the documentation in the method itself:
```java
    /**
     *
     * @return The query string, without the leading ?
     */
    public String getQueryString() {
```
Since `setQueryString()` incorrectly allowed an illegal leading `?` to be set in the query string, it allowed this bad test to pass.  Now that I have fixed the behavior in the exchange to enforce the proper query string (with no leading `?`), the test needs fixed to what it should have been all along.
```java
Assert.assertEquals("URI: /relative/foo relative: /foo QS:bar=a&woz=b bar: a woz: b", response);
```
If you guys are in agreement, I'll fix the test.

bdw429s:
@fl4via @stuartwdouglas I have fixed the bad test I discussed above, however, several of the builds are still failing for what appears to be unrelated issues.  Please let me know if I need to do anything further on this pull.

fl4via:
@bdw429s  thanks for your PR! Those failures are known and, in fact, unrelated.

fl4via (approved):

PR #1173, opened 2021-06-16
[UNDERTOW-1869] fix in memory session creation concurrency

baranowb:

Under review. DO NOT MERGE.

### DIFF
--- a/core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
+++ b/core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
@@ -30,6 +30,7 @@ import java.math.BigInteger;
 import java.math.MathContext;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


baranowb (commented):
Im still tinkering with it. Though Im not entirely sure if it can be improved. Problem with InMemorySessionManager is that at least 3 resources are being used concurrently:
- config(SessionConfig)
- sessions map
- ID generator - which theoretically can produce existing ID. At least code guards against it. Not to mention incoming ID might be "regenerated" by it as well - in theory.

So Im not sure if there is some neat low impact way of doing this.


EDIT: out of curiosity  made a little test, first result is synced, second plain(sample of 15k invocations):
65.53153333333333 ns
63.9204 ns

65.60613333333333 ns
65.37286666666667 ns

65.10233333333333 ns
63.34406666666667 ns

Though, if set to run on 35k, it sometimes reverses  avg invocation time:
58.503457142857144 ns
66.88566666666667 ns

core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
```diff
@@ -56,7 +57,7 @@
 
     private final SessionIdGenerator sessionIdGenerator;
 
-    private final ConcurrentMap<String, SessionImpl> sessions;
```
baranowb, line 59: This change might not be required. ConcurrentMap still might be better since it does not throw concurrent modification exception.

baranowb:
I updated v2, its thinner and should not impact performance as much, but still fix the problem.

fl4via:

PR #1174, opened 2021-06-16
[UNDERTOW-1869] fix concurrency on session creation

baranowb:
Alternative to: https://github.com/undertow-io/undertow/pull/1173 (closed)
Followed up by #1196

### DIFF
--- a/core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
+++ b/core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
@@ -52,6 +52,9 @@ import org.xnio.XnioWorker;
  */
 public class InMemorySessionManager implements SessionManager, SessionManagerStatistics {
 
+    // place holder for sessions during id creation, it prevents collision of session ids in race condition scenarios where two sessions are being created with the same id
+    private static final SessionImpl PLACE_HOLDER_SESSION = new SessionImpl(null, null, null, null, null, null, 0);
+
     private final AttachmentKey<SessionImpl> NEW_SESSION = AttachmentKey.create(SessionImpl.class);
 
     private final SessionIdGenerator sessionIdGenerator;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
```diff
@@ -161,39 +164,49 @@ public Session createSession(final HttpServerExchange serverExchange, final Sess
...
-            int count = 0;
-            while (sessionID == null) {
-                sessionID = sessionIdGenerator.createSessionId();
-                if (sessions.containsKey(sessionID)) {
+            sessionID = sessionIdGenerator.createSessionId();
+        } else {
+            idRequested = true;
+        }
+        SessionImpl session;
+        synchronized(this.sessions) {
```
aogburn, line 176: Would this sync really fix the exception thrown under racing requests?  Seems they'd race up to this sync point, then the first one would create the requested sessionid, then all others would proceed one by one with the exception.

But should we explore a means of fixing this without adding a synchronized block instead?  I recently saw a scenario hitting this migrating from EAP 6 to 7 discussed further here at https://issues.redhat.com/browse/EAPSUP-583.  That particular app showed as much as 20-30 concurrent static requests from a single client, which due to a recent session invalidation and subsequent form auth requirements on those requests, were all racing through InMemorySessionManager.createSession around the same time.  I can imagine such an added sync block would bog those down some.  Not a typical or likely optimal scenario, but probably more of a worst case that we should have in mind if adding such a sync block.

I looked back at EAP 6/JBossWeb, where there was never such an issue complaint.  JBossWeb's ManagerBase.add would allow all the requests to race through without an exception, each one just overwriting the last one in the session map so the last created one ultimately is only preserved.  Looking at Tomcat's ManagerBase, it looks like it still handles it the same way (https://github.com/apache/tomcat/blob/6a667943c5da6b5d61ac6bec1d7c9de061e3217c/java/org/apache/catalina/session/ManagerBase.java#L709) and I'm not aware of any complaints of this approach.  So should we even consider an exception truly required to be thrown up to the app in this case of racing requests with a requested sessionid or just let them race and keep the last session object?

Another option may be keeping InMemorySessionManager the same and letting it throw the ISE.  Instead, try to fix this by catching the IllegalStateException in io.undertow.servlet.spec.ServletContextImpl.getSession's manager.createSession call and trying to get the session at least once more upon catching that exception?

core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
```diff
@@ -145,6 +145,9 @@ public Session createSession(final HttpServerExchange serverExchange, final Sess
                 while (sessions.size() >= maxSize && !evictionQueue.isEmpty()) {
 
                     String key = evictionQueue.poll();
+                    if(key == null) {
```
fl4via, line 148: goo catch!

core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
fl4via: @aogburn I agree that this would not solve the problem. The cause of this lies on this commit #872 . I want to try to skip the exception entirely, reuse the variable and see if: the tests pass, the reproducer from UNDERTOW-1677 passes. If it all passes, IMO this is the way to go: reusing the session ID.

fl4via:

core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
fl4via: @aogburn  and @baranowb  I decided we have two issues at hand: one that is making the session creation safe from the concurrency pov, and the second one is getting rid of the IllegalStateException when users try to reuse the same session. The first one is UNDERTOW-1869 and is solved by this PR.

I created UNDERTOW-1971 for the second issue, and I'm going to update Aaron's PR #1196 with that. That one will be merged for 2.3.0.Final, I will hold from adding it to a minor release as it changes (enhances) the behavior. We are planning 2.3.0.Final for soon, all commits marked with a feature request label will go in that release. Notice that, as I put in the UNDERTOW-1971, this one will be backported to maintenance branch, as there is a request for the fix in that branch.

PR #1180, opened 2021-06-25
[UNDERTOW-1906] Guard against null resource path

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1906
2.0.x PR: https://github.com/undertow-io/undertow/pull/1182

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
@@ -296,7 +296,7 @@ public class ServletContextImpl implements ServletContext {
 
     @Override
     public URL getResource(final String path) throws MalformedURLException {
-        if (!path.startsWith("/")) {
+        if (path == null || !path.startsWith("/")) {
             throw UndertowServletMessages.MESSAGES.pathMustStartWithSlash(path);
         }
         Resource resource = null;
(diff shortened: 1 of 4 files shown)


servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
```diff
@@ -296,7 +296,7 @@ public String getMimeType(final String file) {
 
     @Override
     public URL getResource(final String path) throws MalformedURLException {
-        if (!path.startsWith("/")) {
+        if (path == null || !path.startsWith("/")) {
```
ropalka, line 299: According to javadoc if path argument is null, we should return null.

ropalka (changes requested):

servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
baranowb: This is part of "inconsistency" of this API that I mentioned last week.
JDOC: "The path must begin with a / and is interpreted " - null does not start with "/".

ropalka: Got it.

ropalka (approved):

PR #1183, opened 2021-06-25
[UNDERTOW-1905] remove constant multiplication of timeout in Sesssion…

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1905
2.0.x PR: #1185

### DIFF
--- a/core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
+++ b/core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
@@ -413,7 +413,7 @@ public class InMemorySessionManager implements SessionManager, SessionManagerSta
             this.worker = worker;
             this.evictionToken = evictionToken;
             creationTime = lastAccessed = System.currentTimeMillis();
-            this.maxInactiveInterval = maxInactiveInterval;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
```diff
@@ -413,17 +413,17 @@ private SessionImpl(final InMemorySessionManager sessionManager, final String se
             this.worker = worker;
             this.evictionToken = evictionToken;
             creationTime = lastAccessed = System.currentTimeMillis();
-            this.maxInactiveInterval = maxInactiveInterval;
+            this.setMaxInactiveInterval(maxInactiveInterval);
```
baranowb, line 416: Initially I defined another var to store this, but this might introduce race? Removing unnecessary multiplication in bumpTimeout should be enough.

ropalka (approved):
LGTM

core/src/main/java/io/undertow/server/session/InMemorySessionManager.java
fl4via: agreed

PR #1195, opened 2021-07-06
[UNDERTOW-1359] - fix concurrent change/access to buffer

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1359

Needs perf check/comparison. If impact is too big, most likely some hacky way to handle concurrent access( retain value in local var and work on it + trap exceptions thrown and abort ops if those happen?)


fl4via:
Because synchronizing is not the solution, I'm taking the freedom to close this PR in favor of the other, but I'm suggesting a different change altogether. You might want to just close that PR and submit a new one, @baranowb , so that history does not get lost along the way in case we need to revisit the bug and the proposed fixes

PR #1197, opened 2021-07-07
[UNDERTOW-1359] - fix concurrent change/access to buffer

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1359|
Other PR: https://github.com/undertow-io/undertow/pull/1195

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSourceChannel.java
@@ -122,21 +122,31 @@ public abstract class AbstractFramedStreamSourceChannel<C extends AbstractFramed
             return 0;
         }
         try {
+            final PooledByteBuffer localData = data;
             if (frameDataRemaining == 0 && anyAreSet(state, STATE_LAST_FRAME)) {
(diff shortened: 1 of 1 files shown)


fl4via:
@baranowb  I'm marking alternative PRs as duplicat just to point out that we have more than one PR for the same Jira.

fl4via (changes requested):
There is a simpler alternative IMHO to this. Just set the data to null before invoking close on the points the buffer is cleared. If then we still hear back from someone complaining of the error, we'll consider copying the buffer and synchronizing more. I'll be adding detailed feedback in the Jira

fl4via (changes requested):
@baranowb  I gave this some more thought, and I can see why my suggestion is bogus. 
That being said, I also think that copying does not work at all, because copying to a local variable won't prevent another thread from closing the buffer. Just catching the exception seems to be the way to go here. 
Unless you find I'm mistaken again. In that case, please speak up :-)

baranowb:
Its not about closing, rather than null reference and NPE. If other thread close buffer, we will just get IllegalStateException and thats fine.

ropalka (approved):

fl4via:
I am reviewing this one in light of the latest changes done when fixing the most recent CVE

PR #1198, opened 2021-07-07
[UNDERTOW-1913]apply ReadTimeoutStreamSourceConduit/WriteTimeoutStrea…

ffang:
…mSinkConduit also on the UndertowClient side if READ_TIMEOUT/WRITE_TIMEOUT options are set


baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1913

grgrzybek:
:+1:

fl4via:
This PR introduces a failure in HttpClientTestCase,I'm investigating to see how to fix it

fl4via:
I think it should be fixed now

fl4via:
The failure persists in Windows JDK11. I'm rerunning to see if this is intermittent.

fl4via:
As we can see, the test still fails sporadically. The reason for this is that we are depending on the order the threads run to not get the response. I'll dig deeper to see if there is a workaround, or else we will have to take that away :-/

fl4via:
The test failures have finally been solved. I'm running one last time before merging this PR and doing the releases.

fl4via (approved):

PR #1199, opened 2021-07-12
UNDERTOW-1902 Undertow allows session creation and session ID change after response is committed.

pferraro:
https://issues.redhat.com/browse/UNDERTOW-1902

@fl4via Please review.  I can backport to the relevant branches once you are satisfied.

When integrating into WildFly, we will want to expose this configuration via the ServletContainer management model.

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/UndertowServletMessages.java
+++ b/servlet/src/main/java/io/undertow/servlet/UndertowServletMessages.java
@@ -244,4 +244,10 @@ public interface UndertowServletMessages {
 
     //@Message(id = 10066, value = "Can not set invoke 'declareRoles' from dynamic listener in servlet context for context path '%s' in deployment '%s' ")
     //UnsupportedOperationException cantCallFromDynamicListener(String deploymentName, String contextPath);
+
+    @Message(id = 10067, value = "Servlet container does not permit session creation after response was committed.")
+    IllegalStateException sessionCreationAfterResponseCommittedNotAllowed();
+
+    @Message(id = 10068, value = "Servlet container does not permit session identifier change after response was committed.")
+    IllegalStateException sessionIdChangeAfterResponseCommittedNotAllowed();
 }
--- a/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
+++ b/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
@@ -111,6 +111,7 @@ public class DeploymentInfo implements Cloneable {
     private boolean sendCustomReasonPhraseOnError = false;
     private boolean useCachedAuthenticationMechanism = true;
     private boolean preservePathOnForward = true;
+    private boolean allowOrphanSession = false;
     private AuthenticationMode authenticationMode = AuthenticationMode.PRO_ACTIVE;
     private ExceptionHandler exceptionHandler;
     private final Map<String, ServletInfo> servlets = new HashMap<>();
@@ -1406,6 +1407,14 @@ public class DeploymentInfo implements Cloneable {
         return deploymentCompleteListeners;
     }
 
+    public boolean isOrphanSessionAllowed() {
+        return this.allowOrphanSession;
+    }
+
+    public void setOrphanSessionAllowed(boolean allowOrphanSession) {
+        this.allowOrphanSession = allowOrphanSession;
+    }
+
     @Override
     public DeploymentInfo clone() {
         final DeploymentInfo info = new DeploymentInfo()
@@ -1502,6 +1511,7 @@ public class DeploymentInfo implements Cloneable {
         info.containerMinorVersion = containerMinorVersion;
         info.deploymentCompleteListeners.addAll(deploymentCompleteListeners);
         info.preservePathOnForward = preservePathOnForward;
+        info.allowOrphanSession = allowOrphanSession;
         return info;
     }
 
--- a/servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
@@ -381,7 +381,7 @@ public final class HttpServletRequestImpl implements HttpServletRequest {
             if (!this.servletContext.getDeployment().getDeploymentInfo().isOrphanSessionAllowed()) {
                 throw UndertowServletMessages.MESSAGES.sessionIdChangeAfterResponseCommittedNotAllowed();
             }
-            UndertowServletLogger.REQUEST_LOGGER.sessionIdChangeAfterResponseCommitted();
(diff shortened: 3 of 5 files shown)


fl4via:
@pferraro The overall idea looks good to me. So, IIUC, this is not super urgent because there is a workaround fix in WF ( https://github.com/wildfly/wildfly/pull/14355 ), is that correct? If so, I would like to add this to Undertow 2.3.0.Final, a release I'll be working on within the next weeks (I'll be releasing Undertow 2.2.9.Final within the next days)

servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
```diff
@@ -374,6 +375,12 @@ public String changeSessionId() {
         if (session == null) {
             throw UndertowServletMessages.MESSAGES.noSession();
         }
+        if (this.exchange.getAttachment(ServletRequestContext.ATTACHMENT_KEY).getServletResponse().isCommitted()) {
+            if (!this.servletContext.getDeployment().getDeploymentInfo().isOrphanSessionAllowed()) {
+                throw UndertowServletMessages.MESSAGES.sessionIdChangeAfterResponseCommittedNotAllowed();
+            }
+            UndertowServletLogger.REQUEST_LOGGER.sessionIdChangeAfterResponseCommitted();
```
msfm, line 382: This log is still logged at WARN level even if allowOrphanSession is set to true. So, if a user's application is not compliant with the servlet spec, they can start to see either many IllegalStateException or many WARN log messages after migrating to a newer version. 
- allowOrphanSession == false -> IllegalStateException is thrown.
- allowOrphanSession == true -> WARN message is logged.

I think this message should be logged at DEBUG level (or logged at WARN level only once) if allowOrphanSession == true.

servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
```diff
@@ -888,6 +888,13 @@ public HttpSessionImpl getSession(final ServletContextImpl originalServletContex
                 exchange.putAttachment(sessionAttachmentKey, httpSession);
             } else if (create) {
 
+                if (exchange.getAttachment(ServletRequestContext.ATTACHMENT_KEY).getServletResponse().isCommitted()) {
+                    if (!this.deployment.getDeploymentInfo().isOrphanSessionAllowed()) {
+                        throw UndertowServletMessages.MESSAGES.sessionCreationAfterResponseCommittedNotAllowed();
+                    }
+                    UndertowServletLogger.REQUEST_LOGGER.sessionCreationAfterResponseCommitted();
```
msfm, line 895: Same to the above.

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
```diff
@@ -121,6 +121,7 @@
     private boolean sendCustomReasonPhraseOnError = false;
     private boolean useCachedAuthenticationMechanism = true;
     private boolean preservePathOnForward = true;
+    private boolean allowOrphanSession = false;
```
msfm, line 124: I think the default value should be `true` for compatibility reasons at least during EAP 7.x releases. (I guess it's too late to add a new configuration option to the undertow subsystem for EAP 7.x as it requires the management model change.)
We may be able to change this to `false` for future releases (or set this option to `false` by default in the undertow subsystem layer since EAP 8.x).

pferraro:
@fl4via 
> this is not super urgent because there is a workaround fix in WF ( wildfly/wildfly#14355 ), is that correct?

The workaround exists for WF's distributed session manager only, where the issue is more urgent. For this to result in a memory leak using the in-memory session manager, an application would have to both:
1. Attempt to create a session after a call to HttpServletResponse.sendRedirect(...) or sendError(...).
2. Configure the session with no timeout (either by default <session-timeout>0</session-timeout> or via HttpSession.setMaxInactiveInterval(0))

It is highly improbable that an application would satisfy both conditions, thus the low urgency.

> If so, I would like to add this to Undertow 2.3.0.Final

Sounds good.

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
pferraro: @msfm This proposal is not targeting an existing branch.

pferraro: @msfm By design, Undertow should set its defaults to conform to the servlet specification.  Any container integrating Undertow (e.g. WildFly) is free to set its own default value.

servlet/src/main/java/io/undertow/servlet/spec/HttpServletRequestImpl.java
pferraro: @msfm Worst case scenario, the log frequency would be once per completed user session.
That said, given that an application would have explicitly enabled this, I agree, it is fair to log at a lower level.

servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
msfm: > This proposal is not targeting an existing branch.
>
> By design, Undertow should set its defaults to conform to the servlet specification. Any container integrating Undertow (e.g. WildFly) is free to set its own default value.

I agree. 

My concern was happening to merge the same change into the undertow branches for EAP 7.x. Of course, I understand that the WildFly/EAP integration layer (undertow subsystem layer) can set its own default value. However, I thought having a different default value may cause confusion. 

Anyway, as this change is not targeting existing branches but for future releases, I'm fine with `false` by default.

pferraro:
@fl4via Is the main branch at a state where this PR can be merged?

pferraro:
@fl4via Any update?

pferraro:
@fl4via Any update?

servlet/src/main/java/io/undertow/servlet/UndertowServletMessages.java
```diff
@@ -244,4 +244,10 @@ public interface UndertowServletMessages {
 
     @Message(id = 10066, value = "Can not set invoke 'declareRoles' from dynamic listener in servlet context for context path '%s' in deployment '%s' ")
     UnsupportedOperationException cantCallFromDynamicListener(String deploymentName, String contextPath);
+
+    @Message(id = 10067, value = "Servlet container does not permit session creation after reponse was committed.")
```
fl4via, line 248: I just noticed this reponse instead of response typo. I'm updating the PR so we can finally merge it!

fl4via:
> @fl4via Any update?

@pferraro it is finally being merged. Thanks for waiting for such a long time. As we are better organizing the PR process, the ci configs and splitting tasks among the team, plus moving on to a new feature merge process withing the next months, this hopefully won't happen again.

pferraro:
Great news! Thanks @fl4via !

PR #1205, opened 2021-07-18
[UNDERTOW-1950] Fix NullPointerException in PreCompressedResourceSupplier

TBlueF:
This PR fixes a `NullPointerException` that can occur when using a `PreCompressedResourceSupplier` and trying to access a file where only the compressed version exists.

**For example:**
Currently, if there is a file `data.json.gz` without the uncompressed version `data.json`, then a GET request to that file `data.json` results in a NullPointerException and a `500`-response:
```
java.lang.NullPointerException: Cannot invoke "io.undertow.server.handlers.resource.Resource.getContentType(io.undertow.util.MimeMappings)" because "this.val$originalResource" is null
	at io.undertow.server.handlers.resource.PreCompressedResourceSupplier$1.getContentType(PreCompressedResourceSupplier.java:121)
	at io.undertow.server.handlers.resource.ResourceHandler$1.handleRequest(ResourceHandler.java:299)
	at io.undertow.server.Connectors.executeRootHandler(Connectors.java:370)
	at io.undertow.server.HttpServerExchange$1.run(HttpServerExchange.java:830)
```

With this PR, supplying a pre-compressed resource is no longed dependant on the original resource.

This is achieved by trying to resolve the MIME-Type on our own, instead of asking the original-resource. 
*This might introduce some minor assumptions on the type of the resource, so that would be something to look out for when reviewing this PR :)*

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/resource/PreCompressedResourceSupplier.java
+++ b/core/src/main/java/io/undertow/server/handlers/resource/PreCompressedResourceSupplier.java
@@ -54,20 +54,19 @@ public class PreCompressedResourceSupplier implements ResourceSupplier {
 
     @Override
     public Resource getResource(HttpServerExchange exchange, String path) throws IOException {
-        Resource originalResource = resourceManager.getResource(path);
         if(exchange.getRequestHeaders().contains(Headers.RANGE)) {
             //we don't use serve pre compressed resources for range requests
-            return originalResource;
+            return resourceManager.getResource(path);
         }
-        Resource resource = getEncodedResource(exchange, path, originalResource);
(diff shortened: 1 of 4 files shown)


baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1950

fl4via:
@TBlueF  Can you please provide the NPE stack so we can put in the Jira?

fl4via:
> @TBlueF Can you please provide the NPE stack so we can put in the Jira?

Never mind, I just see the stack trace in the description. I'm trying to reproduce this bug and check if it is possible to write a test for it.

fl4via (approved):

PR #1212, opened 2021-07-19
feat: added two getter methods in DeploymentInfo class

BalbinoOrtus:
add to DeploymentInfo class the getter methods getFilterUrlMapping() and getFilterServletNameMapping() returning
lists with their corresponding data

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
+++ b/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java
@@ -409,9 +409,17 @@ public class DeploymentInfo implements Cloneable {
         return this;
     }
 
+    public List<FilterMappingInfo> getFilterUrlMappings(){
+        return filterUrlMappings;
+    }
+
     public DeploymentInfo addFilterServletNameMapping(final String filterName, final String mapping, DispatcherType dispatcher) {
         filterServletNameMappings.add(new FilterMappingInfo(filterName, FilterMappingInfo.MappingType.SERVLET, mapping, dispatcher));
         return this;
+    }
+
+    public List<FilterMappingInfo> getFilterServletNameMappings(){
+        return filterServletNameMappings;
     }
 
     public DeploymentInfo insertFilterUrlMapping(final int pos, final String filterName, final String mapping, DispatcherType dispatcher) {


BalbinoOrtus:
https://issues.redhat.com/projects/UNDERTOW/issues/UNDERTOW-1930?filter=allissues

BalbinoOrtus:
@fl4via @stuartwdouglas

BalbinoOrtus:
@fl4via I had some errors on the Undertow CI workflow run but they dont seem to be related to the change we did, let me know if there is any problem or error to be fixed

baranowb:
@BalbinoOrtus  could you please edit description and add:
"Issue: https://issues.redhat.com/browse/UNDERTOW-1949"

BalbinoOrtus:
@baranowb let me know if its now ok

fl4via:
@BalbinoOrtus I'm marking it as new feature because it changes signature of API, meaning it is being planned for 2.3.0.Final.
Can you please rebase the PR against current upstream/master? We have fixed the majority of the test failures or disabled the problematic ones, so it will be easier to run the CI check after the rebase.
Last but not least, the thing we need to consider is if we want to provide directly the inner collections or if we should be providing copy of those collections for protection of the inner data against corruption.

bdw429s:
@fl4via To be clear, this pulls **adds** to the API, it does not **change** any existing parts of it.  This is important as it means it is not breaking compatibility.

As to your other question-- yes we specifically need access to the underlying data.  This is because we use Undertow in a dynamic environment, where more than one `web.xml` file can be read-- a default one our tool provides out of the box, and a user-provided one that can override or replace the default servlets, mappings, etc already in the deployment info that we provide.  In this case, we have the ability for the override to remove existing deployment information.  This requires direct access to the underlying Maps and Lists such that we can clear them.  Note, this all happens prior to the Undertow deployment being started.  We don't try to change any of the data after the server has actually started.  Now, that said, I wouldn't mind if there were dedicated methods provide on the `DeploymentInfo` class to clear out servlets, mappings, welcome files, error pages, etc, etc.  But as it stands, there are no methods for removing, so we simply do it like so:
```js
info.getListeners().clear();
```
Of course, there is no way to directly access the filter url and name mapping objects, thus this pull :)

BalbinoOrtus:
@fl4via we have now rebased this PR against current upstream/master, let me know if everything looks ok.

bdw429s:
@fl4via Any updates here? This is still a blocking issue for us and I'm hitting it again in my client work today.  With no programmatic access to the servlet mappings in the deployment info, I can't remove/override servlet mappings on the fly as part of my deployment.

bdw429s:
@fl4via @stuartwdouglas This pull is 2 years old now, why is this taking so long just to merge something that does nothing more than add 2 getter methods to a class?  We've done everything we were asked to do, but  still can't get movement here.

fl4via (changes requested):
I am assuming that the method getFilterMappings in DeploymentInfo cannot be used for the goal you have in mind, am I right?
In that case, I ask yo uto return a copy just like done in DeploymentInfo.getfilterMappings(). The returned collection should not be the same collection kept internally for safety reasons.

fl4via:
> @fl4via To be clear, this pulls **adds** to the API, it does not **change** any existing parts of it. This is important as it means it is not breaking compatibility.
> 
> As to your other question-- yes we specifically need access to the underlying data. This is because we use Undertow in a dynamic environment, where more than one `web.xml` file can be read-- a default one our tool provides out of the box, and a user-provided one that can override or replace the default servlets, mappings, etc already in the deployment info that we provide. In this case, we have the ability for the override to remove existing deployment information. This requires direct access to the underlying Maps and Lists such that we can clear them. Note, this all happens prior to the Undertow deployment being started. We don't try to change any of the data after the server has actually started. Now, that said, I wouldn't mind if there were dedicated methods provide on the `DeploymentInfo` class to clear out servlets, mappings, welcome files, error pages, etc, etc. But as it stands, there are no methods for removing, so we simply do it like so:
> 
> ```js
> info.getListeners().clear();
> info.getFilters().clear();
> etc...
> ```
> 
> Of course, there is no way to directly access the filter url and name mapping objects, thus this pull :)

I see @bdw429s . In that case, I think it makes more sense to have a clearFilterMappings method in DeploymentInfo. @BalbinoOrtus , can you please update the PR with this change? Thank you!

PR #1192, opened 2021-07-01
[UNDERTOW-1910] Make sure that AbstractFramedStreamSinkChannel.awaitW…

fl4via:
…ritable does not block, and use write timeout config for that

Jira: https://issues.redhat.com/browse/UNDERTOW-1910
2.0.x PR: #1217

PR #1217, opened 2021-07-20
[UNDERTOW-1910] Make sure that AbstractFramedStreamSinkChannel.awaitW…

fl4via:
…ritable does not block, and use write timeout config for that

Jira: https://issues.redhat.com/browse/UNDERTOW-1910
Master PR: #1192

### DIFF
--- a/core/src/main/java/io/undertow/UndertowMessages.java
+++ b/core/src/main/java/io/undertow/UndertowMessages.java
@@ -639,4 +639,6 @@ public interface UndertowMessages {
     @Message(id = 203, value = "Invalid ACL entry")
     IllegalArgumentException invalidACLAddress(@Cause Exception e);
 
+    @Message(id = 204, value = "Out of flow control window: no WINDOW_UPDATE received from peer within %s miliseconds")
+    IOException noWindowUpdate(long timeoutMiliseconds);
 }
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/undertow/UndertowMessages.java
```diff
@@ -623,4 +623,7 @@
 
     @Message(id = 199, value = "Read timed out after %s milliseconds.")
     ReadTimeoutException readTimedOut(long timeoutMilliseconds);
+
+    @Message(id = 200, value = "Out of flow control window: no WINDOW_UPDATE received from peer within %s miliseconds")
+    IOException noWindowUpdate(long timeoutMiliseconds);
 }
```
msfm, line 629: It appears this message-id is not the same as [the PR for the master branch](https://github.com/undertow-io/undertow/pull/1192/files#diff-a503bdb32421a0a08548181d1aad3671f05712fadac9ddb0f732013e4163a0ef). 
I think the message-id should be consistent in every release.

fl4via: @msfm  thanks for pointing that out! I'll cherry-pick the fix from master to make sure they both contain the same ids.

fl4via: Okay, done, I cherry-picked the fix from master. :-)

PR #1225, opened 2021-07-27
[UNDERTOW-1932] - unify SSLContext manipulation flow

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1932

Decided to throw exceptions and not silently log as debug - since one of methods already failed like that. Maybe its wrong, but I cant see it.


fl4via:
@baranowb  it looks okay to me. Let's wait on the CI to check if it passes the tests

PR #1233, opened 2021-08-05
[UNDERTOW-1929] Increase even more the sockettimeout config to make s…

fl4via:
…ure that the sporadic failures we are seeing are not an indication the number is still too low


fl4via:
The goal of this PR is to use it to run several times the CI, to make sure that the failures we are seeing are not being caused by lack of a big enough so timeout.

PR #1239, opened 2021-09-02
[UNDERTOW-1357] deprecate Request/TransportProtocolAttribute

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1357


ropalka (approved):

msfm:
The updated change looks good, but you seem to have forgotten to update the commit message (and the title of PR).

fl4via (changes requested):
Please, update the commit message, and add a comment in the @Deprecated annotation saying that RequestProtocolAttribute should be the first choice.

core/src/main/java/io/undertow/attribute/TransportProtocolAttribute.java
```diff
@@ -24,7 +24,9 @@
  * The request method
  *
  * @author Stuart Douglas
+ * @deprecated Please use {@link RequestProtocolAttribute}
  */
+@Deprecated(forRemoval = true)
```
fl4via, line 29: We must discuss when is the best moment to drop support to jdk8. For now, we use jdk8, and the project unfortunately does not compile with this change

fl4via (changes requested):

core/src/main/java/io/undertow/attribute/TransportProtocolAttribute.java
ropalka: Yep, forRemoval annotation attribute have been introduced since JDK9.

fl4via (approved):

fl4via:
this PR is ready to go when we create branch 2.3.0.Final. As discussed with the rest of the team, we'll improve the way of handling feature requests and after that feat request PRs won't be hold anymore.

PR #1244, opened 2021-09-14
[UNDERTOW-1962] Add a StopServerWithExternalWorkerUtils, and make sur…

fl4via:
…e that the tests are waiting for servers to completely stop before starting a new one at the same address

Jira: https://issues.redhat.com/browse/UNDERTOW-1962


fl4via:
Assuming we don't have any more address already in use after this run, I'll replace the latest commit by a new utils, probably called StopServerUtils

PR #1245, opened 2021-09-15
UNDERTOW-1964 IPAddressAccessControlHandler stops working when ProxyPeerAddressHandler is enabled and X-Forwarded-For request header contains multiple IP addresses

msfm:
Issue: 
https://issues.redhat.com/browse/UNDERTOW-1964
https://issues.redhat.com/browse/UNDERTOW-1965
https://issues.redhat.com/browse/UNDERTOW-1975

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/ProxyPeerAddressHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/ProxyPeerAddressHandler.java
@@ -70,9 +70,9 @@ public class ProxyPeerAddressHandler implements HttpHandler {
         if (forwardedFor != null) {
             String remoteClient = mostRecent(forwardedFor);
             //we have no way of knowing the port
-            if(IP4_EXACT.matcher(forwardedFor).matches()) {
+            if (IP4_EXACT.matcher(remoteClient).matches()) {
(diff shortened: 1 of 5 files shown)


msfm:
Hmm, a new test case IPAddressAccessControlHandlerWithProxyPeerAddressHandlerTestCase fails when testing with `-Pproxy` option. 
It appears `setReuseXForwarded(true)` is not invoked for `ProxyHandler` when it's initialized in `DefaultServer`. Hence, as the proxy does not forward the `X-Forwarded-For` request header which the client sends, IPAddressAccessControlHandlerWithProxyPeerAddressHandlerTestCase fails.
I think we need to update `DefaultServer` to initialize `ProxyHandler` with `setReuseXForwarded(true)`, or specify `@ProxyIgnore` for IPAddressAccessControlHandlerWithProxyPeerAddressHandlerTestCase temporarily.

msfm:
Hmm, the test case is still failing with ipv6... I guess the failure is same to [UNDERTOW-1198](https://issues.redhat.com/browse/UNDERTOW-1198 "IPAddressAccessControlHandler does not recognize IPV6 loopback address ::1"), so I hope this failure will be resolved after https://github.com/undertow-io/undertow/pull/1160 is merged.

fl4via:
@msfm it appears that we still have failures in the testsuite

ropalka (approved):

PR #1252, opened 2021-09-28
[UNDERTOW-1969] Ignore deprecated certificate retrieval method failure for deprecated parts of Undertow

ropalka:
https://issues.redhat.com/browse/UNDERTOW-1969

### DIFF
--- a/core/src/main/java/io/undertow/server/BasicSSLSessionInfo.java
+++ b/core/src/main/java/io/undertow/server/BasicSSLSessionInfo.java
@@ -67,7 +67,15 @@ public class BasicSSLSessionInfo implements SSLSessionInfo {
             int i=0;
             for(java.security.cert.Certificate cert : certCol) {
                 this.peerCertificate[i] = cert;
-                this.certificate[i++] = X509Certificate.getInstance(cert.getEncoded());
+                try {
+                    this.certificate[i] = X509Certificate.getInstance(cert.getEncoded());
+                } catch (CertificateException ce) {
+                    // [UNDERTOW-1969] We don't care about deprecated JDK methods failure caused by the fact newer JDKs
+                    // doesn't support them anymore. "this.certificate" is used only by deprecated method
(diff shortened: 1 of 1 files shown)


fl4via (approved):

core/src/main/java/io/undertow/server/BasicSSLSessionInfo.java
```diff
@@ -67,7 +67,15 @@ public BasicSSLSessionInfo(byte[] sessionId, String cypherSuite, String certific
...
             for(java.security.cert.Certificate cert : certCol) {
                 this.peerCertificate[i] = cert;
-                this.certificate[i++] = X509Certificate.getInstance(cert.getEncoded());
+                try {
+                    this.certificate[i] = X509Certificate.getInstance(cert.getEncoded());
+                } catch (CertificateException ce) {
+                    // [UNDERTOW-1969] We don't care about deprecated JDK methods failure caused by the fact newer JDKs
+                    // doesn't support them anymore. "this.certificate" is used only by deprecated method
+                    // {@link SSLSessionInfo.getPeerCertificateChain()} which call should be avoided by API users.
+                    this.certificate[i] = null;
```
bstansberry, line 76: @ropalka @fl4via This is not a big deal but if this happens shouldn't the result be that this.certificate (the whole array) gets set to null? Then getPeerCertificateChain() will throw SSLPeerUnverifiedException which seems cleaner than returning an array with some null elements.

bstansberry: In case you agree I sent up https://github.com/undertow-io/undertow/pull/1255

ropalka: I agree.

fl4via: @bstansberry  thanks for fixing this!

PR #1285, opened 2022-01-06
UNDERTOW-2015 URLUtils QueryStringParser#parse() should not add a que…

msfm:
…ry parameter to HttpServerExchange when parameter name and value are empty

https://issues.redhat.com/browse/UNDERTOW-2015


ropalka (approved):
LGTM

PR #1286, opened 2022-01-10
Fix Nullpointer [UNDERTOW-2017]

md-exerp:


carterkozak:
I believe the null check shouldn't be necessary, but avoids an NPE when expected data is not present. I've run into this problem as well and have pushed a reproducer and (I think) a fix to #1287. The reproducer requires modifications to the HttpRequestConduit, so it shouldn't be merged, but I think that branch makes the problem more obvious (and allows me to hit this NPE every couple runs without the fix applied).

fl4via:
I'm closing this in favor of #1288. If the bug reappears in future releases, please let us know and we'll reopen.

PR #1287, opened 2022-01-13
UNDERTOW-2017 debugging 

carterkozak:
It appears that transitions between states fail to store
relevant data -- this is difficult to reproduce locally because
tests use loopback or otherwise fast networks. By introducing
small buffers and writes that often make no progress, we can
reproduce issues from slower networks.

### DIFF
--- a/core/src/main/java/io/undertow/client/http/HttpRequestConduit.java
+++ b/core/src/main/java/io/undertow/client/http/HttpRequestConduit.java
@@ -280,11 +280,6 @@ final class HttpRequestConduit extends AbstractStreamSinkConduit<StreamSinkCondu
                             do {
                                 res = writeNext(buffer);
                                 if (res == 0) {
-                                    this.string = string;
-                                    this.headerName = headerName;
-                                    this.charIndex = charIndex;
-                                    this.valueIterator = valueIterator;
-                                    this.nameIterator = nameIterator;
                                     log.trace("Continuation");
                                     return STATE_HDR_EOL_CR;
                                 }
@@ -297,11 +292,6 @@ final class HttpRequestConduit extends AbstractStreamSinkConduit<StreamSinkCondu
                             do {
                                 res = writeNext(buffer);
                                 if (res == 0) {
-                                    this.string = string;
-                                    this.headerName = headerName;
-                                    this.charIndex = charIndex;
-                                    this.valueIterator = valueIterator;
-                                    this.nameIterator = nameIterator;
                                     log.trace("Continuation");
                                     return STATE_HDR_EOL_LF;
                                 }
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/undertow/conduits/FixedLengthStreamSourceConduit.java
```diff
@@ -257,7 +257,10 @@ public int read(final ByteBuffer dst) throws IOException {
                 return res = next.read(dst);
             }
         } catch (IOException | RuntimeException | Error e) {
-            IoUtils.safeClose(exchange.getConnection());
+            HttpServerExchange exchangeSnapshot = exchange;
+            if (exchangeSnapshot != null) {
+                IoUtils.safeClose(exchangeSnapshot.getConnection());
+            }
```
carterkozak, line 263: I observed another NPE here, I will cherry-pick this into a separate PR

core/src/test/java/io/undertow/client/http/HttpClientTestCase.java
```diff
@@ -435,4 +456,8 @@ public void failed(IOException e) {
         };
     }
 
+    public static ByteBufferPool getBufferPool() {
+        return new DefaultByteBufferPool(true, 50, 0, 0, 100);
```
carterkozak, line 460: Small buffers result in more frequent write attempts

core/src/main/java/io/undertow/client/http/HttpRequestConduit.java
```diff
@@ -277,8 +278,11 @@ private int processWrite(int state, final ByteBuffer userData) throws IOExceptio
...
                             buffer.flip();
                             do {
-                                res = next.write(buffer);
+                                res = writeNext(buffer);
                                 if (res == 0) {
+                                    this.headerName = headerName;
+                                    this.valueIterator = valueIterator;
+                                    this.nameIterator = nameIterator;
                                     log.trace("Continuation");
                                     return STATE_HDR_EOL_CR;
```
carterkozak, line 287: The transitions to `STATE_HDR_EOL_LF` and `STATE_HDR_EOL_CR` require additional data.

I might suggest saving all state prior to every transition to avoid this kind of bug in the future, however it would require more operations per invocation. That said, in the common case I'd expect the request line and all headers to fit into a single buffer, so it's likely there's no real cost.

fl4via:
Thanks for sharing all this background for testing of UNDERTOW-2017 fix @carterkozak !

core/src/main/java/io/undertow/conduits/FixedLengthStreamSourceConduit.java
carterkozak: Linking that PR here for posterity: #1289

fl4via:
Closing this PR that was created for analysis and reproducing of the UNDERTOW-2017. The bug fix, at #1288, is merged.

PR #1288, opened 2022-01-14
UNDERTOW-2017: HttpRequestConduit state is correctly maintained

carterkozak:
Previously, small buffers that required multiple events to write
would not save the correct header state, resulting in NPEs
and requests which failed to include all headers.

This PR removes the duplicated state between the method body
and HttpRequestConduit in favor of exclusively using
HttpRequestConduit fields. This way the behavior across state
changes is unlikely to be impacted at all.
Hotspot isn't able to optimize field access quite as well
as local method scoped variables, however this isn't likely
to have an impact in practice, and it's more important to
send all headers with every request.


carterkozak:
Tested with #1287

fl4via (approved):

PR #1290, opened 2022-01-15
UNDERTOW-2019: Blocking exchanges can set HEAD response length

carterkozak:
Jira: https://issues.redhat.com/browse/UNDERTOW-2019

### DIFF
--- a/core/src/main/java/io/undertow/io/UndertowOutputStream.java
+++ b/core/src/main/java/io/undertow/io/UndertowOutputStream.java
@@ -26,6 +26,7 @@ import java.nio.channels.FileChannel;
 import io.undertow.UndertowMessages;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.util.Headers;
+import io.undertow.util.Methods;
 import org.xnio.Buffers;
 import io.undertow.connector.ByteBufferPool;
 import io.undertow.connector.PooledByteBuffer;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/io/UndertowOutputStream.java
```diff
@@ -328,10 +329,14 @@ public void close() throws IOException {
         try {
             state |= FLAG_CLOSED;
             if (anyAreClear(state, FLAG_WRITE_STARTED) && channel == null) {
-                if (buffer == null) {
-                    exchange.getResponseHeaders().put(Headers.CONTENT_LENGTH, "0");
-                } else {
-                    exchange.getResponseHeaders().put(Headers.CONTENT_LENGTH, "" + buffer.position());
+                // Head requests may not write to the response, setting content-length directly instead
+                if (!(Methods.HEAD.equals(exchange.getRequestMethod())
+                        && exchange.getResponseHeaders().contains(Headers.CONTENT_LENGTH))) {
```
carterkozak, line 334: Alternatively we could use this path any time `Content-Length` is already set regardless of the method, although that could impact existing code if values are set incorrectly. What do you think?

berler: It looks like now this condition duplicates the isHeadRequestWithContentLength method.

carterkozak: 🤦 thanks! Looks like I pushed too soon!

core/src/test/java/io/undertow/server/handlers/HeadBlockingExchangeTestCase.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
fl4via, line 3: thanks for your PR, @carterkozak !
Can you please change the Copyright year to 2022?

fl4via (changes requested):

core/src/test/java/io/undertow/server/handlers/HeadBlockingExchangeTestCase.java
carterkozak: Updated! Thanks for the review

fl4via (approved):

PR #1292, opened 2022-01-25
[UNDERTOW-1934] invoke onClose in case of network failure

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1934
Ref: https://github.com/undertow-io/undertow/pull/1293

### DIFF
--- a/core/src/main/java/io/undertow/websockets/core/WebSocketChannel.java
+++ b/core/src/main/java/io/undertow/websockets/core/WebSocketChannel.java
@@ -24,6 +24,7 @@ import io.undertow.server.protocol.framed.FrameHeaderData;
 import io.undertow.websockets.extensions.ExtensionFunction;
 import org.xnio.ChannelExceptionHandler;
 import org.xnio.ChannelListener;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/undertow/websockets/core/WebSocketChannel.java
```diff
@@ -82,6 +83,8 @@
      */
     private final Set<WebSocketChannel> peerConnections;
 
+    //TODO: possibly make it i18n msg ?
```
fl4via, line 86: Yes, please add the message to WebSocketMessages, following the same that is done for WebSocketMessages.messageToBig

fl4via (changes requested):
Do you think it would be possible to have a test for this one?

core/src/main/java/io/undertow/websockets/core/WebSocketChannel.java
baranowb: Ok

fl4via (approved):

PR #1299, opened 2022-02-17
[UNDERTOW-2027] Spurious buffer leak error in the test-suite (HttpClientTestCase.testReadTimeout)

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-2027

AFAICT there is no real leak in `HttpClientTestCase.testReadTimeout`. The problem is with the TS. Two problems fixed in this PR related to that test:

1. Re-adding the read timeout when doing a `resumeReads` if not previously set. Added a new test `testReadTimeoutAtInit` which forces the timeout before writing anything back to the client.
2. Fixing how the buffer leaks are detected in the TS. Now the method waits 20s to get a `isEmpty` in the set. The same local variable is always used to avoid that random buffer allocations triggers spurious errors. This took me a lot of time. :smile: 

Thanks!

### DIFF
--- a/core/src/main/java/io/undertow/conduits/ReadTimeoutStreamSourceConduit.java
+++ b/core/src/main/java/io/undertow/conduits/ReadTimeoutStreamSourceConduit.java
@@ -212,6 +212,18 @@ public final class ReadTimeoutStreamSourceConduit extends AbstractStreamSourceCo
     }
 
     @Override
+    public void resumeReads() {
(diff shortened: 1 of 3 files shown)


fl4via (approved):

fl4via:
I'm marking this as a bug fix as well because of the resumeReads in ReadTimeoutStreamSourceConduit.

PR #1302, opened 2022-02-25
[UNDERTOW-2041] Allow leading zeroes for ipv6 addresses

Nowheresly:


Nowheresly:
This pull request removes the check of leading zeroes for ipv6 addresses. According to RFC 4291, it is not required to remove leading zeroes, hence, an ipv6 address with leading zeroes should be considered as valid. See https://datatracker.ietf.org/doc/html/rfc4291 section 2.2

Nowheresly:
Any insight about this PR @stuartwdouglas  ?

rhusar:
Just adding a clickable link: https://issues.redhat.com/browse/UNDERTOW-2041 (please always do that in PR description).

rhusar (approved):
The changes look correct to me; i.e. leading zeros are optional as per RFC.

Nowheresly:
Hi @rhusar . Thanks for taking the time to review my PR. 

Actually, yes. This issue is fairly easy to reproduce with, say Keycloak 16.1.1 (but I guess any application powered by undertow should have the same issue). Just curl any url of Keycloak with a header containing an Ipv6 address with leading zeroes like that: `X-Forwarded-For: 2001:1db8:0100:0003:6:ff00:0042:8329`, and you'll end up with an error 500 returned by undertow...

Admittedly, it is not quite common to have a network device returning a non optimized IPv6 (i.e. with leading zeroes), yet, since we encountered this issue, we came to the conclusion that we cannot reasonably blame the client side for being non-optimized and that's the reason why we proposed this PR.

rhusar:
Right, the reproducer is clear with user supplied IPv6 address. But anyway, @fl4via is the one that with access to the merge button :-)

fl4via (approved):

fl4via (approved):

fl4via:
For some reason, github was not showing the run tests button to me. I had to force push an update to see the tests running.

Nowheresly:
Thanks @fl4via  👍

PR #1304, opened 2022-03-25
[UNDERTOW-2028] Buffer leak error involving AjpServerResponseConduit

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-2028

Changing `AjpServerResponseConduit` to manage responses bigger than the buffer size. Now if the conduit cannot write more to the network it stores several buffers to accommodate the remaining data. If any weird error happens the buffers are closed properly.

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/ajp/AjpServerResponseConduit.java
+++ b/core/src/main/java/io/undertow/server/protocol/ajp/AjpServerResponseConduit.java
@@ -42,8 +42,10 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.FileChannel;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import static org.xnio.Bits.allAreClear;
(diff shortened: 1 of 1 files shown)


fl4via (approved):

core/src/main/java/io/undertow/server/protocol/ajp/AjpServerResponseConduit.java
```diff
@@ -266,6 +268,40 @@ protected void queueCloseFrames() {
         queueFrame(null, buffer);
     }
 
+    private void queueRemainingBytes(final ByteBuffer src, final ByteBuffer[] buffers) {
+        List<PooledByteBuffer> pools = new ArrayList<>(4);
+
+        try {
+            PooledByteBuffer newPooledBuffer = pool.allocate();
```
fl4via, line 275: This buffer is leaking as can be seen here: https://github.com/undertow-io/undertow/runs/5754811783?check_suite_focus=true (windows failure)

core/src/main/java/io/undertow/server/protocol/ajp/AjpServerResponseConduit.java
```diff
@@ -266,6 +268,40 @@ protected void queueCloseFrames() {
...
+
+            ByteBuffer[] savedBuffers = new ByteBuffer[pools.size() + 2];
+            int i = 0;
+            savedBuffers[i++] = buffers[0];
+            for (PooledByteBuffer p : pools) {
+                savedBuffers[i++] = p.getBuffer();
+            }
+            savedBuffers[i] = buffers[2];
+            queueFrame(new PooledBuffersFrameCallback(pools.toArray(new PooledByteBuffer[0])), savedBuffers);
+        } catch (RuntimeException | Error e) {
```
fl4via, line 297: better use a finally block

fl4via (changes requested):

core/src/main/java/io/undertow/server/protocol/ajp/AjpServerResponseConduit.java
rmartinc: Hi @fl4via! 

I think the problem is not at this point. It's what I tried to explain yesterday in the meeting. The buffers here are saved because writes returned 0 (no more writing) and they need to be queued to send them later (when writes are called again). So a finally block to close them is not valid, they should be created now and closed when fully sent. My feeling is that the request is somehow hanging and still on-going in the backend server but the proxy timed out (note the test takes more than 300s, and it usually is around 20s). Other thing is why the test is a success, which is very strange. But it can be a ConnectionClosedException which is ignored [here](https://github.com/undertow-io/undertow/blob/master/servlet/src/test/java/io/undertow/servlet/test/streams/ServletOutputStreamTestCase.java#L271-L275). But I'm not sure about this part.

In summary, I think we have two separate issues involving AjpServerResponseConduit, and this PR just fixes the easy one. We need to investigate what happens later and why the buffer is not closed when sent. It can be because it hangs (maybe related to UNDERTOW-1945 or UNDERTOW-1948) or a missing condition on close (but I have not seen any obvious mistake yet).

What do you think about creating a followup JIRA and at least merging this fix?

fl4via: @rmartinc thanks for the explanation. That sounds fine.

fl4via: @rmartinc  I'm not sure if the failure I saw a couple of days ago was the same. I have just created https://issues.redhat.com/browse/UNDERTOW-2054 to track the Ssl buffer leak errors on Windows proxy-https tests. But I have the impression that, when we were discussing this, we were talking about a buffer created by AjpServerResponseConduit, can yo confirm? As soon as I see that failure again, I'll create the Jira.

fl4via: You can disregard this comment, I found the issue again and I saw your comments in the Jira, here: https://issues.redhat.com/browse/UNDERTOW-2028?focusedCommentId=20027058&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20027058
Let's proceed with UNDERTOW-1945 as it seems it is the same failure indeed. Thanks for having a look at this!

PR #1089, opened 2021-04-09
[UNDERTOW-1879] Utilize o.w.common.Assert class for Null-Checks

boris-unckel:
Fixes https://issues.redhat.com/browse/UNDERTOW-1879

PR #1308, opened 2022-04-20
[UNDERTOW-1879] Utilize o.w.common.Assert class for Null-Checks

boris-unckel:
Fixes https://issues.redhat.com/browse/UNDERTOW-1879


boris-unckel:
@fl4via This is the follow-up on https://github.com/undertow-io/undertow/pull/1089#issuecomment-974329461

PR #1310, opened 2022-04-22
First Round of Jakarta EE 10 Work

ropalka:
This change introduces:
 * minimal Java requirement for Undertow is version 11 since now on
 * removes support for Javax based APIs
 * Jakarta APIs are now default
 * adds WebSocket preview implementation

https://issues.redhat.com/browse/UNDERTOW-2049
https://issues.redhat.com/browse/UNDERTOW-2050
https://issues.redhat.com/browse/UNDERTOW-2051
https://issues.redhat.com/browse/UNDERTOW-2058
https://issues.redhat.com/browse/UNDERTOW-2065


ropalka:
@fl4via please review

servlet/pom.xml
```diff
@@ -29,10 +29,9 @@
     </parent>
 
     <groupId>io.undertow</groupId>
-    <artifactId>undertow-servlet</artifactId>
+    <artifactId>undertow-servlet-jakarta</artifactId>
```
fl4via, line 32: I am not sure it is a good idea to change the module name. I thought we were just going to update the module, and from 2.3.0.Final forward, undertow-servlet uses Jakarta.

websockets-jsr/pom.xml
```diff
@@ -29,11 +29,10 @@
     </parent>
 
     <groupId>io.undertow</groupId>
-    <artifactId>undertow-websockets-jsr</artifactId>
+    <artifactId>undertow-websockets-jsr-jakarta</artifactId>
```
fl4via, line 32: Ditto for the module name change.

fl4via (changes requested):
I really would like to have the websockets and servlet module names unchanged, unless there is a good reason for doing otherwise.

fl4via (commented):
This commit LGTM

fl4via:
@ropalka  in a nutshell, apart from the module name change, all the rest looks good to me

websockets-jsr/pom.xml
ropalka: Fixed. We don't use jakarta suffix anymore.

PR #1316, opened 2022-05-13
Second Round of Jakarta EE 10 Work

ropalka:
This change introduces:

 * Adds Servlet preview implementation

https://issues.redhat.com/browse/UNDERTOW-2052
https://issues.redhat.com/browse/UNDERTOW-2084
https://issues.redhat.com/browse/UNDERTOW-2085
https://issues.redhat.com/browse/UNDERTOW-2086
https://issues.redhat.com/browse/UNDERTOW-2087
https://issues.redhat.com/browse/UNDERTOW-2088
https://issues.redhat.com/browse/UNDERTOW-2089
https://issues.redhat.com/browse/UNDERTOW-2051

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/core/DefaultCharsetMapping.java
+++ b/servlet/src/main/java/io/undertow/servlet/core/DefaultCharsetMapping.java
@@ -24,7 +24,7 @@ import java.util.Properties;
 
 public final class DefaultCharsetMapping {
 
-    public static DefaultCharsetMapping INSTANCE = new DefaultCharsetMapping();
+    public static final DefaultCharsetMapping INSTANCE = new DefaultCharsetMapping();
     private static final String DEFAULT_MAPPING = "/io/undertow/servlet/core/charset.mapping";
     private final Properties map = new Properties();
 


core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -336,6 +336,9 @@ public final class HttpServerExchange extends AbstractAttachable {
      */
     private InetSocketAddress destinationAddress;
 
+    private static final AtomicLong REQUEST_ID_GENERATOR = new AtomicLong(0);
```
boris-unckel, line 339: Is it OK from a security perspective to have predictable ids here? (It's only the pattern that is suspicious, not a fact I know for the case).

core/src/main/java/io/undertow/server/ServerConnection.java
```diff
@@ -46,6 +47,17 @@
  */
 public abstract class ServerConnection extends AbstractAttachable implements ConnectedChannel  {
 
+    private static final AtomicLong CONNECTION_ID_GENERATOR = new AtomicLong(0);
```
boris-unckel, line 50: Is it OK from a security perspective to have predictable ids here? (It's only the pattern that is suspicious, not a fact I know for the case).

core/src/main/java/io/undertow/server/HttpServerExchange.java
fl4via: That's a very good question! From the top of my head, now, it seems it is okay, but I'll do some research before considering this the final implementation for connection ids.

core/src/main/java/io/undertow/server/ServerConnection.java
fl4via: That's a very good question! From the top of my head, now, it seems it is okay, but I'll do some research before considering this the final implementation for connection ids.

PR #1321, opened 2022-05-25
[UNDERTOW-2031] protocol error with HTTP/2 and Expect: 100-continue

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2031


fl4via (approved):

fl4via:
This PR cannot be merged yet, all changes to master are on hold becuase we have everything ready to tag 2.3.0.Alpha1 but nexus is unavailable.

fl4via:
This PR introduced regressions in the tests, some HttpContinue tests are returning 503 (see JDK11 core tests failures)

baranowb:
This seems to happen due to: https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/protocols/http2/Http2FrameHeaderParser.java#L242

Two scenarios are likely here
 - original change does not take into account some corner case and does not set END_STREAM( which would be weird)
 - undertow does not take into account possible :100 in headers as corner case.

fl4via (approved):

PR #1322, opened 2022-05-26
[UNDERTOW-2083] use proper timeout value

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2083


fl4via:
my humble apologies for only noticing now, @baranowb , but this fix duplicates a part of the fix in #1347, even though the error message ended up being a bit different. If I noticed it before, it wouldn't be in my PR. I'm closing this Jira as duplicate and I'll be adding UNDERTOW-2083 to that one. Thanks for your PR anyway!

PR #1325, opened 2022-06-01
[UNDERTOW-2097] DMI_RANDOM_USED_ONLY_ONCE error reported by spotbugs

jstourac:
Let's try to reuse instance of `SecureRandom` class when possible. This
change is attempt to fix error identified by SpotBugs plugin and
described in [1].

[1] https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dmi-random-object-created-and-used-only-once-dmi-random-used-only-once

https://issues.redhat.com/browse/UNDERTOW-2097


jstourac:
just rebased against `master` branch

jstourac:
Failed tests on JDK17 seem to be the very same as are in other PRs so I suppose they are not tied with my changes here.

PR #1333, opened 2022-06-10
[UNDERTOW-2072] DefaultByteBufferPool.getBuffer() may return null

baranowb:
https://issues.redhat.com/browse/UNDERTOW-2072

### DIFF
--- a/core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
+++ b/core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
@@ -259,20 +259,27 @@ public class DefaultByteBufferPool implements ByteBufferPool {
 
         @Override
         public ByteBuffer getBuffer() {
-            if(referenceCount == 0) {
+            final ByteBuffer tmp = this.buffer;
+            //UNDERTOW-2072
+            if (referenceCountUpdater.addAndGet(this, 0) == 0 || tmp == null) {
                 throw UndertowMessages.MESSAGES.bufferAlreadyFreed();
             }
-            return buffer;
+            return tmp;
         }
 
         @Override
         public void close() {
-            if(referenceCountUpdater.compareAndSet(this, 1, 0)) {
-                if(leakDetector != null) {
+            final ByteBuffer tmp = this.buffer;
+            if (referenceCountUpdater.compareAndSet(this, 1, 0)) {
+                this.buffer = null;
+                if (leakDetector != null) {
                     leakDetector.closed = true;
                 }
-                pool.freeInternal(buffer);
-                this.buffer = null;
+                if (tmp != null) {
+                    tmp.position(0);
+                    tmp.limit(0);
+                    pool.freeInternal(tmp);
+                }
             }
         }
 
--- a/core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
+++ b/core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
@@ -139,6 +139,7 @@ public class DefaultByteBufferPool implements ByteBufferPool {
             buffer = queue.poll();
             if (buffer != null) {
                 currentQueueLengthUpdater.decrementAndGet(this);
+                //buffer.clear();
             }
         }
         if (buffer == null) {
@@ -261,7 +262,7 @@ public class DefaultByteBufferPool implements ByteBufferPool {
         public ByteBuffer getBuffer() {
             final ByteBuffer tmp = this.buffer;
             //UNDERTOW-2072
-            if (referenceCountUpdater.addAndGet(this, 0) == 0 || tmp == null) {
+            if (referenceCount == 0 || tmp == null) {
                 throw UndertowMessages.MESSAGES.bufferAlreadyFreed();
             }
             return tmp;
@@ -275,11 +276,7 @@ public class DefaultByteBufferPool implements ByteBufferPool {
                 if (leakDetector != null) {
                     leakDetector.closed = true;
                 }
-                if (tmp != null) {
-                    tmp.position(0);
-                    tmp.limit(0);
-                    pool.freeInternal(tmp);
-                }
+                pool.freeInternal(tmp);
             }
         }
 


e-ts:
Hi @baranowb, are you sure this change actually removes the race? If I understand Java Memory Model correctly, referenceCountUpdate.compareAndSet(0,1) will act as a publish and getBuffers reading 0 will act as acquire. This only provides ordering between actions before publish and actions after acquire. The subsequent write to buffer after publish would still race with the read of buffer, before acquire.

See for example the section JMM Interpretation: Roach Motel on https://shipilev.net/blog/2014/jmm-pragmatics/, and the subsequent quiz examples.

core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
```diff
@@ -256,20 +256,26 @@ private static class DefaultPooledBuffer implements PooledByteBuffer {
...
+                this.buffer = null;
+                if (leakDetector != null) {
                     leakDetector.closed = true;
                 }
-                pool.freeInternal(buffer);
-                this.buffer = null;
+                if (tmp != null) {
+                    pool.freeInternal(tmp);
+                    tmp.position(0);
+                    tmp.limit(0);
```
carterkozak, line 277: I don’t think it’s safe to mutate position and limit after calling freeInternal, the buffer may already be borrowed elsewhere

core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
```diff
@@ -256,20 +256,26 @@ private static class DefaultPooledBuffer implements PooledByteBuffer {
 
         @Override
         public ByteBuffer getBuffer() {
+            final ByteBuffer tmp = this.buffer;
             if(referenceCount == 0) {
```
fl4via, line 260: maybe we should throw exception if tmp is null

fl4via (changes requested):

baranowb:
> Hi @baranowb, are you sure this change actually removes the race? If I understand Java Memory Model correctly, referenceCountUpdate.compareAndSet(0,1) will act as a publish and getBuffers reading 0 will act as acquire. This only provides ordering between actions before publish and actions after acquire. The subsequent write to buffer after publish would still race with the read of buffer, before acquire.
> 
> See for example the section JMM Interpretation: Roach Motel on https://shipilev.net/blog/2014/jmm-pragmatics/, and the subsequent quiz examples.

Oh, Im fairly sure it does not, the problem here is that this part of code wasnt synced for reason and Im sort of forced to accept some trade off. Though Im not 100% happy with even this PR, I will change it a bit.

fl4via (changes requested):
There is just a small change pending, other than that it looks very good @baranowb . I'll add it myself to make it possible adding this commit to the 2.3.0.Beta1 release

core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
fl4via: I'll elaborate a bit more on this. The risk of it being borrowed elsewhere is the same, regardless of whether we clear the buffer or not. If it is borrowed elsewhere and the close is invoked concurrently with the getBuffer mehod, we risk the following order:
* thread 1 gets the buffer
* thread 2 closes the pooled buffer and buffer is made available to a third thread
* thread 3 gets the buffer
* thread 1 uses the buffer
* thread 3 is doomed to see an error when it uses the buffer
So the risk exists regardless.
On the other hand, a normal reusage of the buffer implies that the buffer should be cleared before being reused anyway, so we must clear it here

fl4via: Ok, I just noticed that the clear is being done before returning the pooled byte buffer at allocate. Well, in this case, we can clear it at either one or the other point, either at close or at allocate, the effect will be the same, and in both cases we will have an inconsistency in thread1 or thread3 in the scenario I described above

fl4via (approved):

fl4via:
@baranowb I need to figure out why my proposed commit brokes the tests. I'll update this PR as soon as I'm done with my investigation.

fl4via:
this PR still breaks CI and it will require further investigation

fl4via:
The test failure is:
 Running io.undertow.websockets.extensions.WebSocketExtensionBasicTestCase
[ERROR] Tests run: 3, Failures: 2, Errors: 0, Skipped: 1, Time elapsed: 310.17 s <<< FAILURE! - in io.undertow.websockets.extensions.WebSocketExtensionBasicTestCase
[ERROR] testExtensionsHeaders(io.undertow.websockets.extensions.WebSocketExtensionBasicTestCase)  Time elapsed: 10.094 s  <<< FAILURE!
I canceled the workflow because it seems this causes the CI to hang (or it was just a bad coincidence...)

core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
```diff
@@ -153,7 +154,6 @@ public PooledByteBuffer allocate() {
                 local.allocationDepth++;
             }
         }
-        buffer.clear();
```
ropalka, line 156: This shouldn't be deleted. E.g. thread local cached buffers will not be cleared anymore.

core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
```diff
@@ -139,6 +139,7 @@ public PooledByteBuffer allocate() {
             buffer = queue.poll();
             if (buffer != null) {
                 currentQueueLengthUpdater.decrementAndGet(this);
+                buffer.clear();
```
ropalka, line 142: This doesn't clear all buffers created / reused in this method, see below comment.

ropalka (changes requested):

core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
fl4via: Great catch @ropalka !

fl4via:
I need to see if that fixes the PR so it can be merged. However, it seems that github actions is down, the checks are queued forever. Because I'll be away on PTO until May 2nd, I'm passing the ball to you, @ropalka ! If it passes, it is ready to be merged, unless there are other issues that need to be fixed in the PR. In that case, if someone wants to give it a try, feel free to do so. See you all on the way back!

PR #1339, opened 2022-06-22
[UNDERTOW-2112] Client Cert Renegotiation is not supported by JDK14 and newer. 

kstekovi:
UNDERTOW-2112 Assume the testcase for jdk < 14. New version TLSv1.3 and new JDK versions are breaking this feature

https://issues.redhat.com/browse/UNDERTOW-2112

### DIFF
--- a/core/src/test/java/io/undertow/server/security/ClientCertRenegotiationTestCase.java
+++ b/core/src/test/java/io/undertow/server/security/ClientCertRenegotiationTestCase.java
@@ -34,6 +34,7 @@ import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.entity.StringEntity;
 import org.junit.AfterClass;
+import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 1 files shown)


core/src/test/java/io/undertow/server/security/ClientCertRenegotiationTestCase.java
```diff
@@ -67,7 +67,7 @@ protected List<AuthenticationMechanism> getTestMechanisms() {
 
     @BeforeClass
     public static void startSSL() throws Exception {
-        DefaultServer.startSSLServer(OptionMap.create(SSL_CLIENT_AUTH_MODE, NOT_REQUESTED));
+        DefaultServer.startSSLServer(OptionMap.create(SSL_CLIENT_AUTH_MODE, REQUESTED));
```
kstekovi, line 70: Can be changed SSL_CLIENT_AUTH_MODE to REQUESTED?

jstourac: @kstekovi do you know the reason why this started to be problem with JDK17 and why it works as expected with older JDKs?

jstourac: TBH - I'm not sure whether this change is actually correct. We need more investigation why this behavior changed in JDK17. I've put some comment to the original Jira.

rmartinc:
@fl4via @ropalka Another JDK-17 TS fix. This is a bit more complicated because `force_renegotiation` is starting to make no sense as it's not working in any new version (TLSv1.3 or JDK-14+). The PR assumes the test to be skipped in JDK-14+ (version in which the new extension was enabled by default).

LGTM.

rmartinc (approved):

PR #1340, opened 2022-06-23
[UNDERTOW-2113] Read Timeout Test Case Fail on Windows with JDK17 due to different exception message

kstekovi:
Read Timeout Test Case Fail on Windows with JDK17 due to different exception message

https://issues.redhat.com/browse/UNDERTOW-2113


rmartinc (approved):
LGTM. @fl4via In JDK-17 the error message is different for the exception. It fails all the time in my Windows VM.

fl4via:
marking as enhancement just because this one affects only the test

PR #1341, opened 2022-06-29
[UNDERTOW-2115] Remove size limit to Integer.MAX_VALUE in Receiver#receivePartial*

InkerBot:
Jira: https://issues.redhat.com/browse/UNDERTOW-2115

Javadoc in `io.undertow.io.Receiver#setMaxBufferSize`, "The maximum amount of data to be buffered", and `receivePartial*` won't buffer all `Content-Length`.
So remove this `maxBufferSize` limit in receivePartial*, and add `maxContentSize` limit.
Set `maxContentSize` default value to `Integer.MAX_VALUE`, to sure same behavior with early version without call `Receiver#setMaxContentSize`.
Receiver isn't design for user implementation, it can keep compatibility with early version.


core/src/main/java/io/undertow/io/AsyncReceiverImpl.java
```diff
@@ -498,15 +510,11 @@ public void receivePartialBytes(final PartialBytesCallback callback, final Error
...
-            if (contentLength > Integer.MAX_VALUE) {
-                error.error(exchange, new RequestToLargeException());
-                return;
-            }
         } else {
             contentLength = -1;
         }
-        if (maxBufferSize > 0) {
-            if (contentLength > maxBufferSize) {
+        if(maxContentSize > 0){
```
baranowb, line 516: Im not 100% keen on having another variable doing similar thing.

baranowb: Other than that it looks fine.

ropalka (approved):

core/src/main/java/io/undertow/io/AsyncReceiverImpl.java
fl4via: Agreed.

fl4via (changes requested):
@InkerBot , I think you found an important shortcoming of Undertow that we want to fix. Because I see no point in having two variables for keeping track of contentLength, and because maxBufferSize is not used for defining the maximum size of the buffer, I'll ask you to: remove maxBufferSize field in both receivers; mark setBufferSize() with a Deprecated forRemoval=true; make the setBufferSize edit the maxContentSize field; finally, update javadocs accordingly

fl4via:
@InkerBot I'll ask you to also rebase your PR against main and, if possible, add a test case for it.

I know this PR is very old and we have been struggling to find some time to look at the old PRs... we have finally found the time, but I realize that maybe you don't have any more the time availability to work on this. So, if that's case, please let us know and we will give you a hand in getting the PR in shape for merging.

Thank you!

fl4via:
One interesting idea for this, as a follow up PR, is to add an UndertowOption to set the content size limit.

jasondlee (approved):

PR #1344, opened 2022-07-11
[UNDERTOW-2122] Javax imports in servlet tests

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-2122

There are javax imports (instead of jakarta)  that are creating issues in master. I have fixed that and also added `failIfNoTests` to `true` in the main poms because the error was undetected because of that. If you prefer to not add the last part just let me know.

@ropalka @fl4via Please review this when you have time. This is stopping my other PRs. :smile: 

Thanks!

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -239,6 +239,7 @@
                 <configuration>
                     <enableAssertions>true</enableAssertions>
                     <runOrder>reversealphabetical</runOrder>
+                    <failIfNoTests>true</failIfNoTests>
                     <systemPropertyVariables>
                         <test.ajp>${ajp}</test.ajp>
                         <test.proxy>${proxy}</test.proxy>
--- a/servlet/pom.xml
+++ b/servlet/pom.xml
@@ -197,6 +197,7 @@
                 <configuration>
                     <enableAssertions>true</enableAssertions>
                     <runOrder>reversealphabetical</runOrder>
+                    <failIfNoTests>true</failIfNoTests>
                     <systemPropertyVariables>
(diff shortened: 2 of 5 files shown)


rmartinc:
I have changed my main. I discovered the reason is `failOnError` is `false` in `pom.xml`. It seems that it was done long ago (this [commit](https://github.com/undertow-io/undertow/commit/f382b11fd18b51bdb43407a2acdfe6c780dd31a7)) but better to go revert to the default value and fail if there are compilation errors.

pom.xml
```diff
@@ -185,7 +185,6 @@
                     <configuration>
                         <!-- fork is needed so compiler args can be used -->
                         <fork>true</fork>
-                        <failOnError>false</failOnError>
```
fl4via, line 188: There is a reason why this is set to false. When running all the Undertow tests (like, in your machine) we want to be able to view all failures, and not have the intermittent ones preventing you from seeing, say, a new failure.

fl4via (changes requested):

pom.xml
rmartinc: Then I'm going back to put failIfNoTests to true in servlet, core and websockets folders to detect something weird happens. Give me some minutes.

rmartinc: It's done! Now the surefire plugin fails in the three sub-projects if there are no tests. If there is some error at compilation and the tests are not generated the maven run fails. With this I expect that we will see something strange (tests failures or no test executed) when there are compilation errors.

Thanks!

PR #1347, opened 2022-07-19
[UNDERTOW-2125] [UNDERTOW-2083] Fixes for ReadTimeoutStreamSourceConduit

fl4via:
… connection is not open.

Also:
- prevent unneeded duplicate scheduling by checking if handle is null before rescheduling
- calculate the elapsed time in the expiration properly when throwing the timeout exception

Signed-off-by: Flavia Rainone <frainone@redhat.com>

Jira: https://issues.redhat.com/browse/UNDERTOW-2125
     incorporates https://issues.redhat.com/browse/UNDERTOW-2083

### DIFF
--- a/core/src/main/java/io/undertow/conduits/ReadTimeoutStreamSourceConduit.java
+++ b/core/src/main/java/io/undertow/conduits/ReadTimeoutStreamSourceConduit.java
@@ -18,11 +18,6 @@
 
 package io.undertow.conduits;
 
-import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.nio.channels.FileChannel;
-import java.util.concurrent.TimeUnit;
-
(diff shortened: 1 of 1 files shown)


ropalka (changes requested):
This doesn't look correct from concurrency point of view.

First  'handle' variable should be volatile.
Second either 'double check' or 'atomic updater' should be used to updated handle variable atomically.

Otherwise there are both visibility and atomic assignment issues still present.

fl4via:
@ropalka good point! I made it volatile but then, for some reason, I reverted the change. Now I hope it would be correct. Even though I prefer the atomic field updaters, it wouldn't work here because when we create a new handle we are already rescheduling and we want to do that only when handle is null.

jasondlee (approved):
CI job failure aside, this looks good to me.

PR #1349, opened 2022-07-23
[UNDERTOW-2131] ContentEncodingRepository avoids iterator allocations

schlosna:
Fixes [UNDERTOW-2131](https://issues.redhat.com/browse/UNDERTOW-2131)

`ContentEncodingRepository.getContentEncodings(HttpServerExchange)` allocates many `ArrayList` iterators in hot paths for handling requests. It may be worthwhile to avoid using for-each loops and instead use old fashioned for loop with `List.get(int)` to avoid additional allocations during request processing.
 
```
java.util.ArrayList$Itr
at java.util.ArrayList.iterator()
at java.util.AbstractCollection.addAll(Collection)
at io.undertow.server.handlers.encoding.ContentEncodingRepository.getContentEncodings(HttpServerExchange)
at io.undertow.server.handlers.encoding.EncodingHandler.handleRequest(HttpServerExchange)
 ```

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/encoding/ContentEncodingRepository.java
+++ b/core/src/main/java/io/undertow/server/handlers/encoding/ContentEncodingRepository.java
@@ -50,12 +50,14 @@ public class ContentEncodingRepository {
         }
         final List<EncodingMapping> resultingMappings = new ArrayList<>();
         final List<List<QValueParser.QValueResult>> found = QValueParser.parse(res);
+        //noinspection ForLoopReplaceableByForEach - using induction for loop for iteration to avoid allocation
         for (int i = 0; i < found.size(); i++) {
             final List<QValueParser.QValueResult> result = found.get(i);
             List<EncodingMapping> available = new ArrayList<>();
             boolean includesIdentity = false;
             boolean isQValue0 = false;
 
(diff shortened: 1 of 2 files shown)


schlosna:
@carterkozak curious for your thoughts on this one given how we utilize content encodings.

carterkozak:
@schlosna This change looks reasonable to me, although it may be worth adding a short comment describing why the index-based loop is preferred over a for-each.
For our use-case in particular, I have to wonder if we'd be better off with a custom implementation given we only use a single content-encoding that we know at compile-time. We could avoid all the intermediate collections/sorting/etc.

schlosna:
@carterkozak good call, added comment to suppress IntelliJ `ForLoopReplaceableByForEach` `using induction for loop for iteration to avoid allocation`

fl4via (approved):

schlosna:
@fl4via thanks for the review! I looked at the CI failures and those don't seem related to this change, are there known CI flakes? Is there anything else I need to do before this merges?

Thanks!

fl4via:
@schlosna the failures are related to JDK17, and they came up since we enabled CI with that JDK version. We are currently working on fixing those, rest assured they are not caused by your PR. Thanks for the PR btw!

PR #1353, opened 2022-07-29
UNDERTOW-2112 Client Cert Renegotiation is not supported by JDK14 and newer. 

kstekovi:
UNDERTOW-2112 Assume the testcase for jdk < 14. New version TLSv1.3 and new JDK versions are breaking this feature

https://issues.redhat.com/browse/UNDERTOW-2112

### DIFF
--- a/core/src/test/java/io/undertow/server/security/ClientCertRenegotiationTestCase.java
+++ b/core/src/test/java/io/undertow/server/security/ClientCertRenegotiationTestCase.java
@@ -34,6 +34,7 @@ import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.entity.StringEntity;
 import org.junit.AfterClass;
+import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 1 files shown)


core/src/test/java/io/undertow/server/security/ClientCertRenegotiationTestCase.java
```diff
@@ -67,6 +68,8 @@ protected List<AuthenticationMechanism> getTestMechanisms() {
 
     @BeforeClass
     public static void startSSL() throws Exception {
+        Assume.assumeTrue("UNDERTOW-2112 New version TLSv1.3 and JDK14 and newer versions are breaking this feature",
+                getJavaSpecificationVersion() < 14);
```
carterkozak, line 72: I believe TLSv1.3 was backported to jdk8 a, it may instead be preferable to explicitly specify TLSv1.2 for the server in this test, which should work across java versions

rmartinc: @carterkozak TLSV1.2 is already used in the test but it fails with JDK-17. The problem is that JDK-14+ added a new ticket extension that makes the `force_renegotiation` fail in TLSv1.2. Mainly all new versions (TLS spec of JDK impl) are making `force_renegotiation` fail. See [UNDERTOW-2112](https://issues.redhat.com/browse/UNDERTOW-2112) for more information. The test should be assumed to avoid it in jdk-14+.

rmartinc:
Same PR than https://github.com/undertow-io/undertow/pull/1339 but in 2.2 branch. IMHO `force_renegotiation`is starting to make no sense as it's not working in new versions (TLSv1.3 or JDK-17) and we cannot do anything with it. For the moment the PR just assumes the test to only execute it in JDK versions previous to 14, in which the ticket extension was enabled by default.

rmartinc (approved):

PR #1355, opened 2022-08-01
[UNDERTOW-2138] Removing unnecessary JDK8 reflection hacks - API is available on supported JDKs

ropalka:
https://issues.redhat.com/browse/UNDERTOW-2138


fl4via (approved):

fl4via (changes requested):
I don't think the JDK9 classes are needed either.
The question here, though, is, do we require JDK11 to build or do we require JDK11 to run as well?

ropalka:
The JDK9AlpnProvider class is still necessary on JDK11 and above. Just the class name is unlucky and confusing @fl4via .
I updated this PR with renaming this class from JDK9AlpnProvider to ModularJdkAlpnProvider to better describe it's purpose.

fl4via (approved):

PR #1356, opened 2022-08-02
UNDERTOW-2093 detailed followup.

baranowb:


spotbugs-exclude.xml
```diff
@@ -698,6 +698,45 @@
             <And>
                 <Class name="io.undertow.servlet.core.SessionListenerBridge"/>
                 <Method name="sessionDestroyed"/>
+            </And>
+              <And>
+                <!-- TODO: figure out lambda ref? -->
```
fl4via, line 703: This comment seems to be unrelated to throws runnable rule

core/src/main/java/io/undertow/security/impl/SingleSignOnAuthenticationMechanism.java
```diff
@@ -120,6 +121,8 @@ public void handleNotification(SecurityNotification notification) {
                     });
                     log.tracef("Authenticated account %s using SSO", verified.getPrincipal().getName());
                     return AuthenticationMechanismOutcome.AUTHENTICATED;
+                } else {
```
fl4via, line 124: probably something is missing here?

fl4via (changes requested):

core/src/main/java/io/undertow/security/impl/SingleSignOnAuthenticationMechanism.java
baranowb: AFAIR no

spotbugs-exclude.xml
baranowb: AFAIR lambdas trigger those rules.

PR #1363, opened 2022-08-18
UNDERTOW-2142 ChunkedStreamSinkConduit.write writes all requested buffers

carterkozak:
https://issues.redhat.com/browse/UNDERTOW-2142

### DIFF
--- a/core/src/test/java/io/undertow/server/handlers/ChunkedResponseTransferCodingTestCase.java
+++ b/core/src/test/java/io/undertow/server/handlers/ChunkedResponseTransferCodingTestCase.java
@@ -18,6 +18,7 @@
 
 package io.undertow.server.handlers;
 
+import io.undertow.Handlers;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.server.ServerConnection;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/undertow/server/handlers/ChunkedResponseTransferCodingTestCase.java
```diff
@@ -75,7 +80,36 @@ public void handleRequest(final HttpServerExchange exchange) {
...
+                    }
+                }));
+    }
+
+    @Test
+    public void testMultiBufferWrite() throws IOException {
+        HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/buffers");
+        TestHttpClient client = new TestHttpClient();
+        try {
+            HttpResponse result = client.execute(get);
```
carterkozak, line 107: Without the fix, this test times out as the handler continues attempting to write buffers that ChunkedStreamSinkConduit doesn't drain, as it only looks at the buffer at index 1, where buffers at index 1 and 2 should be written.

ropalka (approved):

fl4via (approved):

PR #1364, opened 2022-08-18
Proposal for a multi-buffer ChunkedStreamSinkConduit.write implementation

carterkozak:
Currently a draft, haven't validated performance differences quite yet!


core/src/main/java/io/undertow/conduits/ChunkedStreamSinkConduit.java
```diff
@@ -128,6 +129,108 @@ public int write(final ByteBuffer src) throws IOException {
...
+        if (totalRemaining > Integer.MAX_VALUE) {
+            // Fall back to one buffer at a time if the total srcs remaining exceeds integer max-value
+            // This should be very rare.
+            for (int i = offset; i < offset + length; i++) {
+                ByteBuffer buf = srcs[i];
+                if (buf.hasRemaining()) {
+                    return write(buf);
+                }
+            }
+            return 0;
```
carterkozak, line 149: In this case, we could internally reduce `length` and potentially dial back the limit on the final buffer, matching the chunkleft>0 case on 170

core/src/main/java/io/undertow/conduits/ChunkedStreamSinkConduit.java
```diff
@@ -214,12 +316,7 @@ public void truncateWrites() throws IOException {
 
     @Override
     public long write(final ByteBuffer[] srcs, final int offset, final int length) throws IOException {
```
carterkozak, line 318: At the cost of some single-element array allocations, we could delegate the single-buffer implementation to this one, just like sslconduit

fl4via (approved):

fl4via (approved):

PR #1365, opened 2022-08-19
UndertowOutputStream writes buffer-sized segments

carterkozak:
Previously, buffers were flushed when exactly filled, or when
a write pushed them over the limit, where the overflow was also
written. Unfortunately this interacts poorly with the chunked
transport encoding sink, which passes each buffer individually
to the delegate such that expensive SSLEngine interactions
are more common than necessary. Concretely, assuming 16k
buffers and 17k writes, each write would result in an array
of a 16k buffer and a 1k buffer, which would be split by
the chunked sink into two calls into the SslConduit, each
with an array of chunking buffer, original data, and chunking
separator buffer.
The new behavior only flushes when requested, or buffers are
full, and the overflow is buffered toward the next write.
It's possible that this may result in poor performance in
cases involving writes many times larger than the buffer
size.

This relates to the investigation which produced #1364 as well


carterkozak:
I wonder if you have thoughts on this, @franz1981 (I was reminded of this change by "[undertow-dev] Sharing performance test results" on the mailing list earlier today :-)

fl4via (approved):

fl4via:
We will need to investigate the failures in Windows before merging this one.

PR #1366, opened 2022-08-23
[UNDERTOW-2082] reassemble cookie crumbs

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2082
This is just temp(for now) for review and discussion.

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/http/HttpRequestParser.java
+++ b/core/src/main/java/io/undertow/server/protocol/http/HttpRequestParser.java
@@ -36,6 +36,8 @@ import io.undertow.util.Methods;
 import io.undertow.util.Protocols;
 import io.undertow.util.URLUtils;
 import io.undertow.util.BadRequestException;
+import io.undertow.util.Cookies;
+
 import org.xnio.OptionMap;
 
 import static io.undertow.util.Headers.ACCEPT_CHARSET_STRING;
@@ -293,9 +295,13 @@ public abstract class HttpRequestParser {
                     handleHeaderValue(buffer, currentState, builder);
                 }
             }
+            sanitazeListTypeHeaders(builder);
+            return;
+        } else {
+            handleStateful(buffer, currentState, builder);
+            sanitazeListTypeHeaders(builder);
             return;
         }
-        handleStateful(buffer, currentState, builder);
     }
 
     private void handleStateful(ByteBuffer buffer, ParseState currentState, HttpServerExchange builder) throws BadRequestException {
@@ -355,6 +361,13 @@ public abstract class HttpRequestParser {
         }
     }
 
+    private void sanitazeListTypeHeaders(final HttpServerExchange builder) {
+        //NOTE: this should be used only for HTTP1.x
+        if(!Cookies.isCrumbsAssemplyDisabled()) {
+            Cookies.assembleCrumbs(builder.getRequestHeaders());
+        }
+        //TODO: list type headers?
+    }
 
     abstract void handleHttpVerb(ByteBuffer buffer, final ParseState currentState, final HttpServerExchange builder) throws BadRequestException;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/undertow/util/Cookies.java
```diff
@@ -417,6 +419,79 @@ private static String unescapeDoubleQuotes(final String value) {
...
+        }
+    }
+
+    /**
+     * IF there is single entry that follows RFC separation rules, it will be turned into singular fields. This should be only
+     * used PRIOR to compression.
+     *
+     * @param headerMap
+     */
+    public static void disperseCrumbs(final HeaderMap headerMap) {
```
fl4via, line 455: Is this a utility method available to Undertow client users? I ask because I don't see any point invoking this method

core/src/test/java/io/undertow/websockets/client/version13/WebSocketClient13TestCase.java
```diff
@@ -60,6 +61,7 @@
  */
 @RunWith(DefaultServer.class)
 @HttpOneOnly
+@Ignore
```
fl4via, line 64: I am assuming you are using this for testing purposes temporarily, for that reason I'm ignoring this for now

core/src/main/java/io/undertow/server/protocol/http/HttpRequestParser.java
```diff
@@ -355,6 +361,13 @@ private void handleStateful(ByteBuffer buffer, ParseState currentState, HttpServ
         }
     }
 
+    private void sanitazeListTypeHeaders(final HttpServerExchange builder) {
+        //NOTE: this should be used only for HTTP1.x
+        if(!Cookies.isCrumbsAssemplyDisabled()) {
+            Cookies.assembleCrumbs(builder.getRequestHeaders());
+        }
+        //TODO: list type headers?
```
fl4via, line 369: Wdy mean here?

fl4via:
@baranowb one point that is calling my attention here for this PR is that, iiuc, the multiple cookie headers can be sent only by HTTP2 clients, is that correct? If that's the case, why edit HttpRequestParser, if it is used for parsing HTTP1 requests?

core/src/test/java/io/undertow/websockets/client/version13/WebSocketClient13TestCase.java
baranowb: Oh my, completely forgot about this, yeah. This test sometimes "randomly fails consistently".

core/src/main/java/io/undertow/util/Cookies.java
baranowb: I did not find example, but RFC wording suggests parts of headers and content can be compressed. If user implements such handler, it can use this method - at least this is how I imagine this pieces should work in undertow?

core/src/main/java/io/undertow/server/protocol/http/HttpRequestParser.java
baranowb: Supposedly there are list type headers - CSV style, though I wasnt able to find any existing example. If such headers are present( multiple instances) those should be squashed into single instance with CSV content.

baranowb:
Should be sent only, but this cant be guaranteed. If client violates specs, currently undertow does nothing. This could warrant 400 response, but this seems like a "soft" error - though, I did not read up/search if there is some general rule. Specs dont mandate 400 in such case, at least in those paragraps.

core/src/main/java/io/undertow/util/Cookies.java
fl4via: Yes, this is how I understood it as well (I was just double checking the intention of the method). Let's just make this clear in the method javadoc in the final version of the PR

core/src/test/java/io/undertow/websockets/client/version13/WebSocketClient13TestCase.java
fl4via: I think this one was fixed by a PR sent by Ricardo.

fl4via: (already merged, the one that solved a TCK issue as well)

fl4via (approved):

PR #1367, opened 2022-08-23
UNDERTOW-2143 Deflaters operate directly on ByteBuffers

carterkozak:
Jira: https://issues.redhat.com/browse/UNDERTOW-2143

### DIFF
--- a/core/src/main/java/io/undertow/conduits/DeflatingStreamSinkConduit.java
+++ b/core/src/main/java/io/undertow/conduits/DeflatingStreamSinkConduit.java
@@ -131,12 +131,22 @@ public class DeflatingStreamSinkConduit implements StreamSinkConduit {
                     return 0;
                 }
             }
+            int initialSrcPosition = src.position();
             int initialRemaining = src.remaining();
-            preDeflate(src);
             deflater.setInput(src);
             Connectors.updateResponseBytesSent(exchange, 0 - initialRemaining);
             deflateData(false);
             int consumed = initialRemaining - src.remaining();
+            int endSrcPosition = src.position();
+            int srcLimit = src.limit();
+            // Reset the buffer to original values with a limit based on what has
+            // been deflated such that only data that has been compressed is
+            // represented by the buffer.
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/conduits/DeflatingStreamSinkConduit.java
```diff
@@ -119,31 +121,35 @@ public int write(final ByteBuffer src) throws IOException {
...
+            if (!src.hasRemaining()) {
                 return 0;
             }
-            //we may already have some input, if so compress it
-            if (!deflater.needsInput()) {
-                deflateData(false);
-                if (!deflater.needsInput()) {
-                    return 0;
-                }
-            }
```
carterkozak, line 131: This case is no longer possible because we deflate from a source buffer and return the difference in position (the amount the deflater has consumed), and set an EMPTY buffer to prevent ourselves from leaking references to that buffer. As such, before and after a call to `write`, the deflater always needs input.

core/src/main/java/io/undertow/conduits/DeflatingStreamSinkConduit.java
```diff
@@ -556,7 +554,6 @@ private void freeBuffer() {
             deflater = null;
             pooledObject.close();
         }
-        IoUtils.safeClose(additionalBuffer);
```
fl4via: I'm missing the points where the buffers are closed? I think you need to close, for example, trailerBuffer here

fl4via (changes requested):

core/src/main/java/io/undertow/conduits/DeflatingStreamSinkConduit.java
carterkozak: `trailerBuffer` is no longer a pooled buffer, rather the result of `ByteBuffer.wrap(` on the byte-array result of `getTrailer()`

fl4via (approved):

PR #1369, opened 2022-08-26
UNDERTOW-2145 OutputStreams only awaitWritable after a write returns zero

carterkozak:
https://issues.redhat.com/browse/UNDERTOW-2145


ropalka (approved):

ropalka (approved):

fl4via (approved):
Just a small observation: I'll update the commit to add an extra space between if and (, just for the sake of code uniformity, but other than this the commit is approved. Good catch!

PR #1372, opened 2022-09-06
[UNDERTOW-2157] UndertowOutputStream: behaviour discrepancies, clarify contract, fix bug

j-baker:
BufferedWriteableOutputStream is currently implemented by:

1. UndertowOutputStream
2. ServletOutputStreamImpl

These have two different behaviours. In UndertowOutputStream, I _believe_ that if the input FileChannel's position is 0, it transfers the whole channel, and if it's non-zero, it crashes.

In ServletOutputStreamImpl, I think the behaviour is something more like:
1. If there is no listener set, transfer the bytes between the current position and end of file.
2. If there is a listener set, transfer the bytes between the current position and end of file, setting the input channel's position as well.

This PR makes the behaviour be more consistent and adds a new, clearer API, deprecating the old one.

1. Clarify the contract of BufferedWriteableOutputStream.transferTo as being the ServletOutputStreamImpl's behaviour, since it has a superset of the UndertowOutputStream's functionality.
2. Never set the position of the input channel in ServletOutputStreamImpl, since it leads to inconsistent behaviour depending on whether the listener is set and seems incongruent with the contract of the underlying FileChannel apis.
3. Add an explicit `transferFrom(FileChannel source, long startPosition, long count)` method to the interface and implement in both implementations.

In theory there is a breaking behaviour change for the user who is relying on their FileChannel's position being set to EOF as a side-effect of the change. Internally to Undertow there is no problem here but obviously this is an API. In my head this is a bugfix but I'd be happy to maintain the present behaviour - I just don't know enough about the 'intended' contract.


servlet/src/main/java/io/undertow/servlet/spec/ServletOutputStreamImpl.java
```diff
@@ -541,35 +546,31 @@ public void transferFrom(FileChannel source) throws IOException {
...
+                pos = startPosition;
 
-                while (size - pos > 0) {
-                    long ret = channel.transferFrom(pendingFile, pos, size - pos);
+                while (remaining - pos > 0) {
+                    long ret = channel.transferFrom(pendingFile, pos, remaining - pos);
                     if (ret <= 0) {
                         clearFlags(FLAG_READY);
                         pendingFile = source;
-                        source.position(pos);
```
j-baker, line 562: this is the piece that seems incongruous

core/src/main/java/io/undertow/io/UndertowOutputStream.java
```diff
@@ -315,10 +320,8 @@ public void transferFrom(FileChannel source) throws IOException {
         if (channel == null) {
             channel = exchange.getResponseChannel();
         }
-        long position = source.position();
-        long size = source.size();
-        Channels.transferBlocking(channel, source, position, size);
-        updateWritten(size - position);
```
j-baker, line 321: these two lines are the present 'actual bug'

fl4via:
thanks for your PR @j-baker ! It will be reviewed shortly and, if approved, added to 2.3.0.Beta2 or 2.3.0.Final.

servlet/src/main/java/io/undertow/servlet/spec/ServletOutputStreamImpl.java
fl4via: yes, I'm not sure why it is doing that. Also, notice that this could throw a NPE, because pendingFile has not been assigned to source

core/src/main/java/io/undertow/io/BufferWritableOutputStream.java
```diff
@@ -34,6 +34,19 @@ public interface BufferWritableOutputStream {
 
     void write(final ByteBuffer byteBuffer) throws IOException;
 
+    /**
+     * Transfer the remaining content of the input FileChannel (from its current position, to end of file).
+     * @deprecated use {@link BufferWritableOutputStream#transferFrom(FileChannel, long, long)}.
+     */
+    @Deprecated
     void transferFrom(FileChannel source) throws IOException;
```
fl4via, line 42: I don't think we should deprecate this method but, instead, add the other method as an alternative.

fl4via (changes requested):

fl4via:
Because we did not get a response in a timely manner, I will be updating this PR to incorporate the feedback so it can be merged.

fl4via:
After somethought, I decided to remove the new method and keep the fixes. The reason for this is that, with this, it can be backported to old branches, and there is no need for a new method. The interface is used internally. If someone raises a need for a new method, then we will of course introduce it.

fl4via (approved):

PR #1378, opened 2022-09-21
UNDERTOW-2162: Only canonicalize the servlet path when getting request dispatcher

YaoFu-ea:
**Issue:** https://issues.redhat.com/browse/UNDERTOW-2162

**Description:** : We found this issue after upgrade to Wildfly 26.1.2, and the following code no longer work as expected
```
RequestDispatcher rd = request.getRequestDispatcher("/aaa?url=http://bb.com");
rd.forward(request, response);
``` 
The forwarded request has path `/aaa?url=http:/bb.com` which caused by canonicalize the whole path with query params in `ServletContextImpl`. 

**Fix**: Only canonicalize the servlet path in `ServletContextImpl`

**Test**:
* Unit test added in `DispatcherForwardTestCase`
* Tested locally with the updated undertow-servlet 2.2.19 jar in wildfly 26.1.2 modules, works great.

### DIFF
--- a/servlet/src/test/java/io/undertow/servlet/test/dispatcher/DispatcherForwardTestCase.java
+++ b/servlet/src/test/java/io/undertow/servlet/test/dispatcher/DispatcherForwardTestCase.java
@@ -255,4 +255,21 @@ public class DispatcherForwardTestCase {
             client.getConnectionManager().shutdown();
         }
     }
+
+    @Test
+    public void testIncludesUrlInPathParameters() throws IOException {
+        TestHttpClient client = new TestHttpClient();
+        try {
+            HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/servletContext/dispatch");
+            get.setHeader("forward", "/path?url=http://test.com");
+            HttpResponse result = client.execute(get);
+            Assert.assertEquals(StatusCodes.OK, result.getStatusLine().getStatusCode());
+            String response = HttpClientUtils.readResponse(result);
+            // Path parameters should not be canonicalized
+            Assert.assertEquals("pathInfo:null queryString:url=http://test.com servletPath:/path requestUri:/servletContext/path", response);
+
+        } finally {
+            client.getConnectionManager().shutdown();
+        }
+    }
 }
--- a/servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java
@@ -343,12 +343,18 @@ public class ServletContextImpl implements ServletContext {
         if (!path.startsWith("/")) {
             throw UndertowServletMessages.MESSAGES.pathMustStartWithSlashForRequestDispatcher(path);
         }
-        final String realPath = CanonicalPathUtils.canonicalize(path, true);
-        if (realPath == null) {
+        int qsPos = path.indexOf("?");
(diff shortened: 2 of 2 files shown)


servlet/src/test/java/io/undertow/servlet/test/dispatcher/DispatcherForwardTestCase.java
```diff
@@ -255,4 +255,21 @@ public void testIncludesPathParameters() throws IOException {
             client.getConnectionManager().shutdown();
         }
     }
+
+    @Test
+    public void testIncludesUrlInPathParameters() throws IOException {
+        TestHttpClient client = new TestHttpClient();
+        try {
+            HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/servletContext/dispatch?url=http://test.com");
+            get.setHeader("forward", "/path;pathparam=foo");
```
rmartinc, line 264: I think that the query parameter with the slash should go to the header, not the initial url. The current test would work in the same way without your fix. Something like this:

```java
HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/servletContext/dispatch?a=b");
 get.setHeader("forward", "/path;pathparam=foo?url=http://test.com");
```

rmartinc (commented):
Thanks a lot for the PR @YaoFu-ea!

I think that the fix is OK but the test is not checking the issue. The query parameter with the slash should go to the request dispatcher path, which is the header (not the initial URL).

Can you modify the test? And test that it fails without your fix and it's solved with it?

Thanks!

servlet/src/test/java/io/undertow/servlet/test/dispatcher/DispatcherForwardTestCase.java
YaoFu-ea: Yep! Good catch!
Changes pushed. ✅ 
Verified the test failed before the fix
Before fix

And the test passed after the fix!

rmartinc: Thanks a lot! It's OK now. Just a nitty request... Do you mind to squash the two commits into a single one?

YaoFu-ea: Done! ✅ 
Wondering if branch 2.2.X still accept changes, if so, I can graft this fix to 2.2.X branch and create a new PR. Thanks!

rmartinc (approved):
@fl4via @ropalka This LGTM. When doing a dispatch the query string was also taken into account for doing the canonicalization. The Fix just removes the query string to do that only in the path. Tests execute OK.

The PR is for master, if you want it in some other branch just let @YaoFu-ea know.

ropalka (approved):

servlet/src/test/java/io/undertow/servlet/test/dispatcher/DispatcherForwardTestCase.java
fl4via: @YaoFu-ea thank you! There is no need, the bug fixes are backported weekly or bi-weekly to 2.2.x

YaoFu-ea: That's great! 👏 Thank you for the info!

PR #1381, opened 2022-10-10
Final Undertow Master Cleanup

ropalka:
Jira: https://issues.redhat.com/browse/UNDERTOW-2168

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -25,11 +25,11 @@
     <parent>
         <groupId>io.undertow</groupId>
         <artifactId>undertow-parent</artifactId>
-        <version>2.3.0.Beta2-SNAPSHOT</version>
+        <version>2.3.0.Final-SNAPSHOT</version>
     </parent>
 
     <artifactId>undertow-benchmarks</artifactId>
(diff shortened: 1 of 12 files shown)


benchmarks/pom.xml
```diff
@@ -28,10 +28,10 @@
         <version>2.3.0.Beta2-SNAPSHOT</version>
     </parent>
 
+    <groupId>io.undertow</groupId>
     <artifactId>undertow-benchmarks</artifactId>
     <version>2.3.0.Beta2-SNAPSHOT</version>
-
-    <name>Undertow Benchmarks</name>
+    <name>Undertow Benchmarks - Jakarta EE Variant</name>
```
fl4via, line 34: I don't see why adding Jakarta EE variant to the pom descriptions, in the future this will seem dated. If the whole project is just Jakarta compatible, we can assume that each module is Jakarta EE compatible

fl4via (changes requested):
We need a Jira for this

benchmarks/pom.xml
ropalka: Fixed @fl4via . I also created JIRA for this PR.

fl4via (approved):

PR #1383, opened 2022-10-10
[UNDERTOW-2167] 2.2.x branch will provide Java EE API version only

ropalka:
https://issues.redhat.com/browse/UNDERTOW-2167


servlet/pom.xml
```diff
@@ -31,8 +31,7 @@
     <groupId>io.undertow</groupId>
     <artifactId>undertow-servlet</artifactId>
     <version>2.2.21.Final-SNAPSHOT</version>
-
-    <name>Undertow Servlet</name>
+    <name>Undertow Servlet - Java EE Variant</name>
```
fl4via, line 34: In the same way, I don't think we should add Java EE variant here. I think we could state something about this on the README file instead

fl4via (changes requested):

servlet/pom.xml
ropalka: Fixed @fl4via

fl4via (approved):

PR #1384, opened 2022-10-11
[UNDERTOW-1818] AbstractServletInputStreamTestCase.runTestParallel fails with bytes out of order

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-1818

There are different changes in the PR although the former is the main one (the rest fixes minor issues in the same `AbstractServletInputStreamTestCase` test):

1. The direct solution for the issue is in the `Connectors` class. As the method is called `ungetRequestBytes` the order is directly changed. If we identify that something needs to insert real new data at the current point in the exchange (not data that was read before and is push back to be read again) another method can be created with the previous order. I think that all the callers are using the re-read / push-back idea.
2. Adding a max queue size because sometimes proxy tests fail (error 500) when concurrency tests start in input stream classes. The proxy conf now allows the requests to be queued up to 20 (max concurrency).
3. Setting a timeout of 60s for all the proxy configurations and inside the async stream tests. Previously it was 30 and 120 (although servlets had default 30s), so aligning everything to 60s. It was enough for me.
4. Reducing the concurrency of the input stream tests to 4 * CPUs up to previous 20 (maximum). Envs with only a few CPUs are limited to cope with the requests in 60s.

In my envs (windows-2016 and linux-fedora) the `ServletInputStreamRequestBufferingTestCase` was executed with `-P proxy` option 50 times without issues.  Let's see how the tests go.

### DIFF
--- a/core/src/main/java/io/undertow/server/Connectors.java
+++ b/core/src/main/java/io/undertow/server/Connectors.java
@@ -144,8 +144,12 @@ public class Connectors {
             System.arraycopy(buffers, 0, newArray, 0, buffers.length);
         } else {
             newArray = new PooledByteBuffer[existing.length + buffers.length];
-            System.arraycopy(existing, 0, newArray, 0, existing.length);
-            System.arraycopy(buffers, 0, newArray, existing.length, buffers.length);
+            // If there are previous buffers we are re-buffering data so although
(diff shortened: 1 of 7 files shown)


fl4via (approved):

fl4via:
Superseds #1003

PR #1390, opened 2022-10-14
[UNDERTOW-2185] Upgrade Apache DS to 2.0.0.AM26

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-2185

Changes:

* Upgrade apacheds to 2.0.0.AM26, now the test artifact should be used in combination with the kerberos interceptor for the spnego tests.
* Tests previously used things from apacheds dependencies, all of them have been removed and substituted by two methods in the `HttpClientUtils` utility test class.
* The dependency for log4j has been removed from autobahn cos I think it's not used for anything in the tests.

Just remind the comment in the Jira. The version 2.0.0.AM26 has an issue with IBM jdk-8 and it doesn't work with it. So take care and do not backport to 2.2.x if the TS should work with that runtime. Last version that works with IBM jdk-8 is AM25.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -111,9 +111,15 @@
         
         <dependency>
             <groupId>org.apache.directory.server</groupId>
-            <artifactId>apacheds-all</artifactId>
+            <artifactId>apacheds-test-framework</artifactId>
             <scope>test</scope>
-        </dependency>        
+        </dependency>
+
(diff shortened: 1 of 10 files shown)


ropalka (approved):

ropalka:
Thanks @rmartinc !

PR #1391, opened 2022-10-18
[UNDERTOW-2186] allow WEB-INF or META-INF subdirectory access

aogburn:
https://issues.redhat.com/browse/UNDERTOW-2186

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/handlers/Paths.java
+++ b/servlet/src/main/java/io/undertow/servlet/handlers/Paths.java
@@ -35,7 +35,7 @@ final class Paths {
     static boolean isForbidden(final String path) {
         final StringTokenizer st = new StringTokenizer(path, "/\\", false);
         String subPath;
-        while (st.hasMoreTokens()) {
+        if (st.hasMoreTokens()) {
             subPath = st.nextToken();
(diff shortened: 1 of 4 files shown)


webfolderio:
No, this shouldn't be allowed. It's not safe to access those paths.

aogburn:
Note that this isn't to allow the top level .war/WEB-INF or .war/META-INF to be accessed.  That is still blocked.  It's to allow other lower application sub directories that might happen to also be named WEB-INF or META-INF (like .war/foo/WEB-INF/page).  That is perhaps unconventional, but they should not be sensitive like the top level .war/(WEB-INF|META-INF).  Things like .war/foo/WEB-INF/page could be accessed before through undertow and for comparison are still accessible through other servlet containers like Tomcat.

fl4via (approved):

fl4via:
I did some research to find the original UNDERTOW-1981 fix #1258 , and I can see that the goal of the original issue was to prevent access being given to the WEB-INF/META-INF directories when there is a difference in the upper/lower case of the letters and when the path is preceded by // for example.
I  can see that @ropalka 's fix is using StringTokenizer, and that one does not return empty tokens, so this fix seems correct to me.

PR #1393, opened 2022-10-26
[UNDERTOW-2054] Buffer leak errors involving SslConduit on CI

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-2054

There are errors with leaks in CI runs involving `SslConduit` (mainly `doUnwrap` method). For example this [run](https://github.com/undertow-io/undertow/actions/runs/3321170261/jobs/5488507673) or this [run](https://github.com/undertow-io/undertow/actions/runs/3235864053/jobs/5300966510). As commented in  the JIRA I think that the reason is that there is a race between the `close` and the `doUnwrap` (called by a handle read in the channel). The `doUnwrap` checks at the beginning that it is not closed but it can be closed during the execution. And if a new `dataToUnwrap` was buffered after the close it is never freed. So just adding a new condition that cleans the buffers in the finally if the conduit is closed. If you prefer to throw a closed exception or similar just let me know (but I decided to just clean it).

Let's see the CI.

### DIFF
--- a/core/src/main/java/io/undertow/protocols/ssl/SslConduit.java
+++ b/core/src/main/java/io/undertow/protocols/ssl/SslConduit.java
@@ -881,8 +881,8 @@ public class SslConduit implements StreamSourceConduit, StreamSinkConduit {
                 requiresListenerInvocation = true;
             }
             if (dataToUnwrap != null) {
-                //if there is no data in the buffer we just free it
-                if (!dataToUnwrap.getBuffer().hasRemaining()) {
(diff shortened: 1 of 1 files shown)


rmartinc:
@fl4via @ropalka Take a look to this PR when you have time. This is the most frequent failure in the CI right now (probably with the `DefaultServletCachingListenerTestCase.testWelcomePages` which I'm also managing in XNIO-415). Don't know if it fixes all the problems but I think that it will solve some of them.

fl4via (approved):

PR #1402, opened 2022-11-07
[UNDERTOW-2198] Add Option to force the SNIHostName

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-2198

Adding new option `SSL_SNI_HOSTNAME` to force the SNI hostname. This is needed by the wildfly-http-client which uses IP to connect but the SNI should be set to teh correct hostname. Tests added that uses hostname/ip to connect to the server (tests are assumed if the address does not resolve to hostname).

PR to master.
PR for 2.2.x: https://github.com/undertow-io/undertow/pull/1403

### DIFF
--- a/core/src/main/java/io/undertow/UndertowOptions.java
+++ b/core/src/main/java/io/undertow/UndertowOptions.java
@@ -325,6 +325,13 @@ public class UndertowOptions {
      */
     public static final Option<Boolean> SSL_USER_CIPHER_SUITES_ORDER = Option.simple(UndertowOptions.class, "SSL_USER_CIPHER_SUITES_ORDER", Boolean.class);
 
+    /**
+     * This option forces {@link io.undertow.protocols.ssl.UndertowXnioSsl} to use a specific
+     * name as the {@link javax.net.ssl.SNIHostName} for a client connection. If the option is
+     * not set, the URL hostname is used when available (for URLs that do not use the IP address).
+     * The option can be useful when connecting via IP but want to set a SNI hostname value.
+     */
+    public static final Option<String> SSL_SNI_HOSTNAME = Option.simple(UndertowOptions.class, "SSL_SNI_HOSTNAME", String.class);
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/undertow/UndertowOptions.java
```diff
@@ -325,6 +325,13 @@ public class UndertowOptions {
      */
     public static final Option<Boolean> SSL_USER_CIPHER_SUITES_ORDER = Option.simple(UndertowOptions.class, "SSL_USER_CIPHER_SUITES_ORDER", Boolean.class);
 
+    /**
+     * This option forces {@link io.undertow.protocols.ssl.UndertowXnioSsl} to use a specific
+     * name as the {@link javax.net.ssl.SNIHostName} for a client connection. If the option is
+     * not set, the hostname in the URL, if available (i.e. not using IP), is used. The option
```
fl4via, line 331: "If the option is not set, the hostname in the URL, if available (i.e. not using IP), is used".
It would look clearer if this were reorded as in:
"If the option is not set, the hosname in the URL is used, if available (i.e., not using IP)."
also, I would replace the second if by a when just to prevent using the same proposition twice in the same statement and take away the "in the":
"If the option is not set, the URL hostname is used when available (for URLs that do not use the IP address)

fl4via (approved):
It looks great! I'm just suggesting a minor improvement to the Javadoc.

core/src/main/java/io/undertow/UndertowOptions.java
rmartinc: Done!

rmartinc:
I'm very happy to see that 2.2.x PR failed with the issue with SslConduit leak buffer and it didn't happen here in master (I have rebased to incorporate the fix for it). :smiley:

fl4via:
@rmartinc me too! It is good to see it fixed!

PR #1403, opened 2022-11-07
[UNDERTOW-2198] Add Option to force the SNIHostName

rmartinc:
Issue: https://issues.redhat.com/browse/UNDERTOW-2198

Adding new option `SSL_SNI_HOSTNAME` to force the SNI hostname. This is needed by the wildfly-http-client which uses IP to connect but the SNI should be set to teh correct hostname. Tests added that uses hostname/ip to connect to the server (tests are assumed if the address does not resolve to hostname).

PR to 2.2.x branch.
PR for master: https://github.com/undertow-io/undertow/pull/1402


fl4via:
@rmartinc we are not adding new options to minor releases. But we can add a system property if this is really needed for this version, is that the case?

rmartinc:
Changes done as in the master PR.

PR #1414, opened 2022-11-29
[UNDERTOW-2200] - Path and query parameters are not decoded properly due to flag switch

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2200
Deprecatred by: https://github.com/undertow-io/undertow/pull/1438

### DIFF
--- a/core/src/test/java/io/undertow/server/EncodedEncodedSlashTestCase.java
+++ b/core/src/test/java/io/undertow/server/EncodedEncodedSlashTestCase.java
@@ -46,28 +46,10 @@ public class EncodedEncodedSlashTestCase {
         });
     }
 
-    @Test
-    public void testSlashNotDecoded() throws Exception {
-
-        final TestHttpClient client = new TestHttpClient();
-        try {
-            HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/%2f%5c");
-            HttpResponse result = client.execute(get);
-            Assert.assertEquals(StatusCodes.OK, result.getStatusLine().getStatusCode());
-            Assert.assertEquals("/%2f%5c", HttpClientUtils.readResponse(result));
-
-        } finally {
-            client.getConnectionManager().shutdown();
-        }
-    }
-
-
     @Test @ProxyIgnore
     public void testSlashDecoded() throws Exception {
 
         final TestHttpClient client = new TestHttpClient();
-        OptionMap old = DefaultServer.getUndertowOptions();
-        DefaultServer.setUndertowOptions(OptionMap.create(UndertowOptions.ALLOW_ENCODED_SLASH, true));
         try {
             HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/%2f%5c");
             HttpResponse result = client.execute(get);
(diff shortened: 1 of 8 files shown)


core/src/test/java/io/undertow/server/EncodedEncodedSlashTestCase.java
```diff
@@ -46,35 +46,42 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
...
+    @Test @ProxyIgnore
+    public void testSlashDecoded() throws Exception {
 
         final TestHttpClient client = new TestHttpClient();
         try {
             HttpGet get = new HttpGet(DefaultServer.getDefaultServerURL() + "/%2f%5c");
             HttpResponse result = client.execute(get);
             Assert.assertEquals(StatusCodes.OK, result.getStatusLine().getStatusCode());
-            Assert.assertEquals("/%2f%5c", HttpClientUtils.readResponse(result));
+            Assert.assertEquals("//\\", HttpClientUtils.readResponse(result));
```
rmartinc, line 57: @baranowb Are you changing the behavior of ALLOW_ENCODED_SLASH? I think that by default ALLOW_ENCODED_SLASH was false, which means that slashes are not decoded and remain as %2F and %5C. After this fix would be the opposite, they are decoded into real slashes by default. Note this was done because of CVE-2007-0450 in tomcat.

baranowb: https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/UndertowOptions.java#L123
IMHO, it should be way more clearly described. This change makes flag work according to description.

ALLOW - means something is permitted, so in this case encoded characters == dont decode those. In test above ALLOW=false, which should force docode on those chars - as per description.

core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -527,6 +527,8 @@ public String getRequestPath() {
      * @param requestPath the request URI path
      */
     public HttpServerExchange setRequestPath(final String requestPath) {
+        if(requestPath!= null && requestPath.equals("//\\"))
+            new Exception().printStackTrace();
```
ropalka, line 531: Please throw IllegalArgumentException instead of Exception plus use logging facility and not system.out.

ropalka (changes requested):

core/src/main/java/io/undertow/server/HttpServerExchange.java
baranowb: Weird, I dont have it in my local branch. Need to check, since I was working from my laptop

carterkozak:
If I understand correctly, this inverts the meaning of `UndertowOptions.ALLOW_ENCODED_SLASH` in a way that will break undertow users when dependency versions are upgraded. If I set `ALLOW_ENCODED_SLASH`, it's generally to the non-default value otherwise there's little point in setting it, and this change will break the behavior that I meant to configure. Perhaps we could solve this problem with better documentation, or alternatively by deprecating the option and creating a new one with a better name?

baranowb:
@carterkozak yeah, thats the problem with this one. I just proposed fix that demonstrate the problem here. How it will end up being solved is still up for debate. No worry, its not being considered due to very reason you mentioned.

carterkozak:
Thanks for the context :-)

fl4via:
> If I understand correctly, this inverts the meaning of `UndertowOptions.ALLOW_ENCODED_SLASH` in a way that will break undertow users when dependency versions are upgraded. If I set `ALLOW_ENCODED_SLASH`, it's generally to the non-default value otherwise there's little point in setting it, and this change will break the behavior that I meant to configure. Perhaps we could solve this problem with better documentation, or alternatively by deprecating the option and creating a new one with a better name?

Yes @carterkozak we are discussing the best alternatives for this. According to the javadoc `ALLOW_ENCODED_SLASH` stands for:
> If a request comes in with encoded / characters (i.e. %2F), will these be decoded.
     * <p>
     * This can cause security problems if a front end proxy does not perform the same decoding, and as a result
     * this is disabled by default.
     * <p>
     * Defaults to false

So, according to that, the `ALLOW_ENCODED_SLASH` property indicates that %2F should be decoded to /. But, the name ALLOW_ENCODED_SLASH implies otherwise, it implies that we should allow the %2F encoded form to pass through without being decoded during request parsing.

Maybe having a new UndertowOption with a better name for 2.4.0.Final is worth considering.

fl4via:
Closing because this PR is deprecated by #1438

PR #1424, opened 2022-12-14
[UNDERTOW-2210] Improved write ASCII path on ServletPrintWriter

franz1981:


franz1981:
Still WIP, I'll run some micro and end 2 end benchmark soon with the final implementation.

franz1981:
@fl4via @stuartwdouglas 
I've added a JMH benchmark (a bit unfair because it doesn't iterate through different strings), which results are (on my machine):

```
Benchmark                      (encoderType)  (inLength)  (outCapacity)  Mode  Cnt     Score    Error  Units
AsciiEncodingBenchmark.encode          batch           7           8196  avgt   20     0.025 ±  0.001  us/op
AsciiEncodingBenchmark.encode          batch          19           8196  avgt   20     0.044 ±  0.001  us/op
AsciiEncodingBenchmark.encode          batch         248           8196  avgt   20     0.255 ±  0.002  us/op
AsciiEncodingBenchmark.encode          batch       12392           8196  avgt   20    12.686 ±  1.038  us/op
AsciiEncodingBenchmark.encode          batch      493727           8196  avgt   20   488.882 ±  2.891  us/op
AsciiEncodingBenchmark.encode        noBatch           7           8196  avgt   20     0.025 ±  0.001  us/op
AsciiEncodingBenchmark.encode        noBatch          19           8196  avgt   20     0.042 ±  0.001  us/op
AsciiEncodingBenchmark.encode        noBatch         248           8196  avgt   20     0.409 ±  0.001  us/op
AsciiEncodingBenchmark.encode        noBatch       12392           8196  avgt   20    17.131 ±  0.133  us/op
AsciiEncodingBenchmark.encode        noBatch      493727           8196  avgt   20   706.596 ±  6.517  us/op
AsciiEncodingBenchmark.encode        vanilla           7           8196  avgt   20     0.026 ±  0.001  us/op
AsciiEncodingBenchmark.encode        vanilla          19           8196  avgt   20     0.052 ±  0.001  us/op
AsciiEncodingBenchmark.encode        vanilla         248           8196  avgt   20     0.546 ±  0.006  us/op
AsciiEncodingBenchmark.encode        vanilla       12392           8196  avgt   20    28.323 ±  0.216  us/op
AsciiEncodingBenchmark.encode        vanilla      493727           8196  avgt   20  1089.093 ± 26.762  us/op
```
I didn't yet spend time into profiling (with perf) to see if bound check elimination has been effective and I still need to run a full fat end 2 end test with SpecJ, but code-wise, should be a in a decent state for a review.

carterkozak:
:wave: Hi there, @franz1981 :-)

Out of curiosity, how do the benchmarks above compare to some variation of `ByteBuffer.put(string.getBytes(StandardCharsets.US_ASCII))`? (relevant blogpost from Claes Redestad: https://cl4es.github.io/2021/10/17/Faster-Charset-Encoding.html -- I realize the allocation will have broader impact, but I'm curious about the comparison point!)

franz1981:
For smaller strings is very likely will perform better then what I have done, but not fully sure; let me do it, I am curious too
And Hi sir @carterkozak  😁 I was sure I would have triggered you eheh

franz1981:
And I didn't tried with varhandle byte buffer view (that will save, likely the reverse bytes due to endianess)

franz1981:
Results from SpecJ are very good: as expected it deliver a reduction of overhead ~2X 
ie from 2.91% to 1.43% of cpu samples presence for the overall test duration and, for the specific stack trace related JSP 
from 51.82% to 31,19% that's still very good. 
I'll double check the ASM of JMH to see if there's anything better thanks to var handle byte buffer view and a comparison with lang encoders as I've promised to @carterkozak

franz1981:
Interestingly, JDK 19 (note: prev results were using JDK 11) seems to disagree:
```
Benchmark                      (encoderType)  (inLength)  (outCapacity)  Mode  Cnt    Score    Error  Units
AsciiEncodingBenchmark.encode          batch           7           8196  avgt   20    0.020 ±  0.001  us/op
AsciiEncodingBenchmark.encode          batch          19           8196  avgt   20    0.041 ±  0.001  us/op
AsciiEncodingBenchmark.encode          batch         248           8196  avgt   20    0.253 ±  0.013  us/op
AsciiEncodingBenchmark.encode          batch       12392           8196  avgt   20   12.771 ±  0.574  us/op
AsciiEncodingBenchmark.encode          batch      493727           8196  avgt   20  499.805 ± 11.393  us/op
AsciiEncodingBenchmark.encode        noBatch           7           8196  avgt   20    0.019 ±  0.001  us/op
AsciiEncodingBenchmark.encode        noBatch          19           8196  avgt   20    0.025 ±  0.001  us/op
AsciiEncodingBenchmark.encode        noBatch         248           8196  avgt   20    0.161 ±  0.004  us/op
AsciiEncodingBenchmark.encode        noBatch       12392           8196  avgt   20    8.856 ±  0.095  us/op
AsciiEncodingBenchmark.encode        noBatch      493727           8196  avgt   20  329.699 ±  1.925  us/op
AsciiEncodingBenchmark.encode        vanilla           7           8196  avgt   20    0.021 ±  0.001  us/op
AsciiEncodingBenchmark.encode        vanilla          19           8196  avgt   20    0.035 ±  0.001  us/op
AsciiEncodingBenchmark.encode        vanilla         248           8196  avgt   20    0.303 ±  0.007  us/op
AsciiEncodingBenchmark.encode        vanilla       12392           8196  avgt   20   19.122 ±  0.039  us/op
AsciiEncodingBenchmark.encode        vanilla      493727           8196  avgt   20  722.313 ± 59.218  us/op
```
and report the first commit of this PR to be the winner; I'll dig a bit more on this

franz1981:
I believe this can be made even faster by reading chars as long via varhandle/methodhandle, will experiment with it next week

servlet/src/main/java/io/undertow/servlet/spec/ServletPrintWriter.java
```diff
@@ -305,6 +376,110 @@ public void write(final String s, final int off, final int len) {
...
+            final char c6 = s.charAt(off + 6);
+            final char c7 = s.charAt(off + 7);
+            if (c7 > 127 || c6 > 127 || c5 > 127 || c4 > 127 ||
+                    c3 > 127 || c2 > 127 || c1 > 127 || c0 > 127) {
+                return i << 3;
+            }
+            final long leBatch = (((long) (c7) << 56) |
+                    ((long) (c6 & 0xff) << 48) |
+                    ((long) (c5 & 0xff) << 40) |
+                    ((long) (c4 & 0xff) << 32) |
```
theRealAph, line 421:
```suggestion
            final long leBatch = (((long) (c7) << 56) |
                    ((long) c6 << 48 >> 8) |
                    ((long) c5 << 48 >> 16) |
                    ((long) c4 << 48 >> 24) |
``
... etc

File: servlet/src/main/java/io/undertow/servlet/spec/ServletPrintWriter.java [hunk 2]
```diff
@@ -305,6 +376,110 @@ public void write(final String s, final int off, final int len) {
...
+                    ((long) (c3 & 0xff) << 24) |
+                    ((long) (c2 & 0xff) << 16) |
+                    ((long) (c1 & 0xff) << 8) |
+                    ((long) (c0 & 0xff)));
+            buffer.putLong(out, leBatch);
```
theRealAph, line 426: Better to mask and shift longs here.

servlet/src/main/java/io/undertow/servlet/spec/ServletPrintWriter.java
```diff
@@ -254,45 +236,86 @@ public void write(final char[] buf, final int off, final int len) {
...
+                return i;
+            }
+        }
+        return i;
+    }
+
+    private static int setAsciiBE(ByteBuffer buffer, int out, char[] chars, int off, int len) {
+        final int longRounds = len >>> 3;
+        for (int i = 0; i < longRounds; i++) {
+            final long batch1 = (((long) chars[off]) << 48) |
```
franz1981, line 264: @theRealAph not very happy about these casts; I'm going to refresh the knowledge of JLS, but without them (or `& 0xFFFFL`)  the chars read were treated as `int` and too large shifts were just truncating/losing data

servlet/src/main/java/io/undertow/servlet/spec/ServletPrintWriter.java
```diff
@@ -254,45 +236,84 @@ public void write(final char[] buf, final int off, final int len) {
...
+                    chars[off + 4] << 16 |
+                    chars[off + 6];
+            final long batch2 = (((long) chars[off + 1]) << 48) |
+                    (((long) chars[off + 3]) << 32) |
+                    chars[off + 5] << 16 |
+                    chars[off + 7];
+            if (((batch1 | batch2) & 0xff80_ff80_ff80_ff80L) != 0) {
+                return i << 3;
+            }
+            final long batch = (batch1 << 8) | batch2;
```
franz1981, line 275: @theRealAph I've removed 

```java
            final long maskedBatch1 = (batch1 & 0x007f_007f_007f_007fL) << 8;
            final long maskedBatch2 = batch2 & 0x007f_007f_007f_007fL;
            final long batch = maskedBatch1 | maskedBatch2;
```
because of the previous check re being  in `[0, 127]`

benchmarks/src/main/java/io/undertow/benchmarks/AsciiEncoders.java
```diff
@@ -0,0 +1,149 @@
...
+            for (int i = 0; i < longRounds; i++) {
+                final long batch1 = (((long) chars[off]) << 48) |
+                        (((long) chars[off + 2]) << 32) |
+                        chars[off + 4] << 16 |
+                        chars[off + 6];
+                final long batch2 = (((long) chars[off + 1]) << 48) |
+                        (((long) chars[off + 3]) << 32) |
+                        chars[off + 5] << 16 |
+                        chars[off + 7];
+                if (((batch1 | batch2) & 0xff80_ff80_ff80_ff80L) != 0) {
```
whitingjr, line 55: To avoid auto-boxing shouldn't the right hand side of the comparison be 0L ?

franz1981:
@fl4via this is now ready for review; it delivers a great speedup (up to 2X) across different JDK versions (11->19)

servlet/src/main/java/io/undertow/servlet/spec/ServletPrintWriter.java
theRealAph: Cast looks correct to me.

franz1981: Nice, thanks!

servlet/src/main/java/io/undertow/servlet/spec/ServletPrintWriter.java
```diff
@@ -254,45 +236,84 @@ public void write(final char[] buf, final int off, final int len) {
         write(cb);
     }
 
+    private static int writeAndFlushAscii(ServletOutputStreamImpl outputStream, ByteBuffer buffer, char[] chars, int start, int end) throws IOException {
+        assert buffer.order() == ByteOrder.BIG_ENDIAN;
+        int i = start;
```
theRealAph, line 241: So I'm curious: why is the buffer big endian?

franz1981: I have noticed that the existing code doesn't change it from the default endianness of NIO, but if your question is re "why it was like that?" I have my own theories:
1. By design: existing framing protocol uses big endian buffers for some by-design reason
2. Historical: networking often makes uses of big endian buffers 
3. Uups: NIO buffers while allocated are big endian due to historical reasons AFAIK, and the existing code doesn't enforce native and/or little endianess because the http txt protocol doesn't make use of it 

I am keen to vote for the third option @stuartwdouglas probably knows

benchmarks/src/main/java/io/undertow/benchmarks/AsciiEncoders.java
```diff
@@ -0,0 +1,200 @@
+package io.undertow.benchmarks;
```
fl4via, line 1: please add the header

benchmarks/src/main/java/io/undertow/benchmarks/AsciiEncodingBenchmark.java
```diff
@@ -0,0 +1,85 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019 Red Hat, Inc., and individual contributors
```
fl4via, line 3: 2023

fl4via (changes requested):

fl4via (approved):

ropalka (approved):

ropalka:
Thanks @franz1981 !

franz1981:
Placing this comment here to remember: https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/lang/StringLatin1.java#L52 shows that whatever single byte char is fine, meaning 0-255 that translate in a wider encoding domain (that's latin ie US-ASCII extended) then what we've used in undertow. TLDR we can still perform fast path ascii encoding up to 255 included (and the check mask of this optimized method can change accordly!)

PR #1408, opened 2022-11-17
UNDERTOW-2194 - improve cookie parsing

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2194
Proposed changes to parse cookies "more properly"

PR #1428, opened 2022-12-14
UNDERTOW-2194 - improve cookie parsing so its aligned with spec without 3rd part deps.

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2194
This will replace: https://github.com/undertow-io/undertow/pull/1408


core/src/main/java/io/undertow/server/MultiValueHashListStorage.java
```diff
@@ -0,0 +1,170 @@
...
+ * This class allows to store multiple values for single key value. Key and value order are honored.
+ * Duplicate entries of key-value pair is not allowed, it is vetted via {@link java.lang.Object#equals()} method.
+ * @author baranowb
+ *
+ * @param <K> - key type
+ * @param <V> - value type
+ */
+public class MultiValueHashListStorage<K, V> {
+    // NOTE: requests are processed by single thread so no need to sync?
+    private LinkedHashMap<K, ArrayList<V>> storage = new LinkedHashMap<>();
```
ropalka, line 38: 'storage' field should have 'final' classifier.

core/src/main/java/io/undertow/server/MultiValueHashListStorage.java
```diff
@@ -0,0 +1,170 @@
...
+
+    public void put(K key, V value) {
+        ArrayList<V> lst = storage.get(key);
+        if (lst == null) {
+            lst = new ArrayList<>();
+            lst.add(value);
+            storage.put(key, lst);
+        } else {
+            if (lst.contains(value)) {
```
ropalka, line 47: It is not effective to remove value with same hash code and then add it again.
The following code does the same job: if (!lst.contains(value)) lst.add(value);

core/src/main/java/io/undertow/server/MultiValueHashListStorage.java
```diff
@@ -0,0 +1,170 @@
...
+            if (lst != null && lst.contains(value)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public ArrayList<V> get(K key) {
+        final ArrayList<V> values = storage.get(key);
+        return values != null ? values : new ArrayList<V>();
```
ropalka, line 90: Use Collections.emptyList() instead of new ArrayList<V>() to prevent creating memory garbage.

core/src/main/java/io/undertow/server/MultiValueHashListStorage.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    public Collection<V> removeAll(Object key) {
+        final ArrayList<V> values = storage.remove(key);
+        return values != null ? values : new ArrayList<V>();
```
ropalka, line 95: Same here use Collections.emptyList() instead of new ArrayList<V>() to prevent creating memory garbage.

core/src/main/java/io/undertow/server/MultiValueHashListStorage.java
```diff
@@ -0,0 +1,170 @@
...
+
+    public Set<K> keySet() {
+        return storage.keySet();
+    }
+
+    public void clear() {
+        storage.clear();
+    }
+
+    private class DefaultIterator<V> implements Iterator<V> {
```
ropalka, line 120: DefaultIterator should have ideally both "static final" classifiers.  With "static" classifier  DefaultIterator class could  be shared among all MultiValueHashListStorage object instances. Without "static" classifier new DefaultIterator class is created per MultiValueHashListStorage instances. This change will result in improvements to shared memory AKA "method area" .

core/src/main/java/io/undertow/util/Cookies.java
```diff
@@ -211,51 +212,80 @@ public static Map<String, Cookie> parseRequestCookies(int maxCookies, boolean al
...
+    public static void parseRequestCookies(int maxCookies, boolean allowEqualInValue, List<String> cookies, MultiValueHashListStorage<String, Cookie> parsedCookies) {
+        parseRequestCookies(maxCookies, allowEqualInValue, cookies, parsedCookies, LegacyCookieSupport.COMMA_IS_SEPARATOR);
+    }
+
     @Deprecated
     static Map<String, Cookie> parseRequestCookies(int maxCookies, boolean allowEqualInValue, List<String> cookies, boolean commaIsSeperator) {
         return parseRequestCookies(maxCookies, allowEqualInValue, cookies, commaIsSeperator, LegacyCookieSupport.ALLOW_HTTP_SEPARATORS_IN_V0);
     }
 
+    @Deprecated(since="2.3.1", forRemoval=true)
```
ropalka, line 229: Since this is public method please add javadoc that will contain at least this line and will guide developers in their migration path:

@deprecated use {@link #parseRequestCookies(int, boolean, List, MultiValueHashListStorage)} instead

core/src/main/java/io/undertow/util/Cookies.java
```diff
@@ -211,51 +212,80 @@ public static Map<String, Cookie> parseRequestCookies(int maxCookies, boolean al
...
-            parseCookie(cookie, parsedCookies, maxCookies, allowEqualInValue, commaIsSeperator, allowHttpSepartorsV0);
+            parseCookies(cookie, parsedCookies, maxCookies, allowEqualInValue, commaIsSeperator, allowHttpSepartorsV0);
         }
 
         final Map<String, Cookie> retVal = new TreeMap<>();
-        for (Cookie cookie : parsedCookies) {
-            retVal.put(cookie.getName(), cookie);
+        Iterator<Cookie> cookiesIterator = parsedCookies.valuesIterator();
+        Cookie cookie = null;
+        for (;cookiesIterator.hasNext(); ) {
```
ropalka, line 251: This is not a problem at all but 'while' would be a better choice than 'for' with empty  'initialization' and 'post action' section. Another option is to move 'Cookie cookie = null' declaration to 'for initialization' section.

ropalka (changes requested):

core/src/main/java/io/undertow/server/OverridableTreeSet.java
```diff
@@ -22,6 +22,7 @@
 /**
  * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
+@Deprecated(since="2.3.1", forRemoval=true)
```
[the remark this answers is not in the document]
baranowb: huh, got fooled by IDE. When I ^H it, it picks up binary deps from other projects and still show class as "in use"...

core/src/main/java/io/undertow/util/Cookies.java
baranowb: I have no idea "why" I did it this way

core/src/main/java/io/undertow/server/MultiValueHashListStorage.java
baranowb: I think there was a reason for this, but I cant remember now.

baranowb: yeah, its required by CookieTest. Im going to debug it, since its fishy.

baranowb:
Ive set "under veirfication" in light of crossreviewing with another RFE. Details in doc.

baranowb:
NOTE: this changes tests to have $Version as first, AFAIR that was expectation from RFC, but not sure if we should or should not enforce, maybe it was done this way by design here, given that some old server/UA might be doing it like that ?

PR #1434, opened 2023-01-21
[UNDERTOW-2231] Test failure is fixed by adding this pull request

java-codehunger:
This pull request will solve the Issue https://issues.redhat.com/browse/UNDERTOW-2231


core/src/main/java/io/undertow/client/http/HttpClientExchange.java
```diff
@@ -122,6 +122,9 @@ void setContinueResponse(ClientResponse response) {
     void setResponse(ClientResponse response) {
         this.response = response;
         if (responseCallback != null) {
+            try {
+                Thread.sleep(500);
```
fl4via, line 126: We can't add a sleep to a core class because of a test, only to the test itself.

fl4via (changes requested):

PR #1448, opened 2023-03-23
[UNDERTOW-2254] Included the HttpServerExchange in the HostSelector.

jeham:
This allows for advanced routing based on eg. header values or attributes.
Jira: https://issues.redhat.com/browse/UNDERTOW-2254

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/proxy/LoadBalancingProxyClient.java
+++ b/core/src/main/java/io/undertow/server/handlers/proxy/LoadBalancingProxyClient.java
@@ -360,7 +360,7 @@ public class LoadBalancingProxyClient implements ProxyClient {
             }
         }
 
-        int host = hostSelector.selectHost(hosts);
+        int host = hostSelector.selectHost(hosts, exchange);
 
         final int startHost = host; //if the all hosts have problems we come back to this one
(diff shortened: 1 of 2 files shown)


fl4via (changes requested):
I'm approving this for Undertow 2.4.0.Final (because of the changes in signatures.
However, we will want to keep the old HostSelector .selectHost methods and mark them as deprecated with forRemoval set to true. 
I'll also ask you to add the [UNDERTOW-2254] prefix to the commit message.
Thank you for your PR!

fl4via:
Also, please rebase it because the test failures we are seeing are supposed to have been fixed in master (UNDERTOW-2252), so hopefully after rebase they will be gone.

fl4via:
> Thanks for the feedback @fl4via ! I updated the code and the commit message. I tried to rebase but got conflicts in all POM files though.

@jeham do you mind if I do the rebase then? I can pull a rebased version and kick off CI.
There is one thing that is calling my attention, though, and when I first reviewed this I didn't take notice. The PR should be done against master.
Do you need this feature specifically for 2.2.x?

jeham:
Thanks for the reply @fl4via - I'll create a new PR from master.

PR #1465, opened 2023-04-13
[UNDERTOW-2216] fix resource serving via predicates + test

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2216

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/resource/DirectoryUtils.java
+++ b/core/src/main/java/io/undertow/server/handlers/resource/DirectoryUtils.java
@@ -19,6 +19,9 @@
 package io.undertow.server.handlers.resource;
 
 import io.undertow.UndertowLogger;
+import io.undertow.predicate.PathPrefixPredicate;
+import io.undertow.predicate.Predicate;
+import io.undertow.predicate.PathPrefixPredicate.PathPrefixMatchRecord;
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/undertow/server/handlers/resource/DirectoryUtils.java
```diff
@@ -90,7 +94,7 @@ public static boolean sendRequestedBlobs(HttpServerExchange exchange) {
         return false;
     }
 
-    public static StringBuilder renderDirectoryListing(String path, Resource resource) {
+    public static StringBuilder renderDirectoryListing(final HttpServerExchange exchange, String path, final Resource resource) {
```
fl4via, line 97: @baranowb can the original method be deprecated with an attribute marking it for removal on 2.4.0.Final?

fl4via (changes requested):

core/src/main/java/io/undertow/server/handlers/resource/DirectoryUtils.java
baranowb: @fl4via yes, however, the question is this fix required, given that more or less this can be achieved with some URL rewriting?

fl4via (approved):

PR #1468, opened 2023-04-14
Remove uses of classes in javax.security.cert

eirbjo:
Efforts are underway in the OpenJDK project to remove the long deprecated-for-removal classes in the package javax.security.cert. These classes were introduced for backwards compatibility concerns with the unbundled JSSE release for JDK 1.2/1.3, but their use have been discouraged since they were introduced.

It would be good to update Undertow to not depend on / use these archaic APIs.

See https://bugs.openjdk.org/browse/JDK-8227024 and the corresponding CSR https://bugs.openjdk.org/browse/JDK-8227395

Changes:

- Remove `SSLSessionInfo.getPeerCertificateChain` and its overrides in `BasicSSLSessionInfo`, `ConnectionSSLSessionInfo` and `Http2SslSessionInfo`
- Remove a catch of `javax.security.cert.CertificateException` in `AjpRequestParseState.createSslSessionInfo`
- Remove the util method `Certificates.toPem` which takes `javax.security.cert.X509Certificate` as parameter and seems unused
- SNISSLEngine.InitalState overrides `SSLEngine.getHandshakeSession` and returns a custom, mostly no-op `SSLSession`.  Since `getHandshakeSession` does not seem to be called, I opted to simply remove this override and the custom SSLSession implementation. This way we also don't need to implement `SSLSession.getPeerCertificateChain` which returns `javax.security.cert.X509Certificate[]`. Please advice if this is correct.

### DIFF
--- a/core/src/main/java/io/undertow/protocols/ssl/SNISSLEngine.java
+++ b/core/src/main/java/io/undertow/protocols/ssl/SNISSLEngine.java
@@ -17,8 +17,6 @@ package io.undertow.protocols.ssl;
 
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
-import java.security.Principal;
-import java.security.cert.Certificate;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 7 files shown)


fl4via:
Thanks @eirbjo ! I am approving this to run CI, but this is not being merged for now. I'll get back from PTO on May, and by then I'll sort out all new feature/API change PRs, they will go into a branch scheduled for 2.4.0.Final.

eirbjo:
> Thanks @eirbjo ! I am approving this to run CI, but this is not being merged for now. I'll get back from PTO on May, and by then I'll sort out all new feature/API change PRs, they will go into a branch scheduled for 2.4.0.Final.

Thanks @fl4via ! This change is not entirely trival, so I'm expecting it can take some careful review. While there is no immediate rush to get this done, I'm very happy that you can take a look at this when you are back at work. Enjoy your time off :-)

fl4via:
@eirbjo  I apologize for the huge delay, we have been very overloaded with many tasks since you submitted this PR.

I kno it has been a long time and I saw the updated status of the Jiras you linked.

We will be running tests and studying the impact of this so we can decide how to proceed with these changes.

PR #1471, opened 2023-04-18
[UNDERTOW-2263] UNDERTOW-2093 detailed followup

baranowb:
Replace https://github.com/undertow-io/undertow/pull/1356
Jira: https://issues.redhat.com/browse/UNDERTOW-2093

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/proxy/ProxyProtocolReadListener.java
+++ b/core/src/main/java/io/undertow/server/protocol/proxy/ProxyProtocolReadListener.java
@@ -103,7 +103,7 @@ class ProxyProtocolReadListener implements ChannelListener<StreamSourceChannel> 
         } catch (IOException e) {
             UndertowLogger.REQUEST_IO_LOGGER.ioException(e);
             IoUtils.safeClose(streamConnection);
-        } catch (Exception e) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/undertow/server/protocol/proxy/ProxyProtocolReadListener.java
```diff
@@ -103,9 +103,6 @@ public void handleEvent(StreamSourceChannel streamSourceChannel) {
         } catch (IOException e) {
             UndertowLogger.REQUEST_IO_LOGGER.ioException(e);
             IoUtils.safeClose(streamConnection);
-        } catch (Exception e) {
```
fl4via, line 106: I don't think we should be excluding this catch @baranowb

fl4via (changes requested):
There is one change that I don't think we should be doing in the PR.

core/src/main/java/io/undertow/server/protocol/proxy/ProxyProtocolReadListener.java
baranowb: hmm. brain fart. for some reason I assumed RE is not tied to E....

fl4via (approved):

PR #1472, opened 2023-04-18
[UNDERTOW-2255] Solve the Hpack decoding exception

fl4via:
…by making sure that proxy handler does not forward http2 settings incompatible with the proxy local settings.

This was causing the proxy to send a bigger header table size to server in the HTTP2-Settings headers. As a consequence, the server would try to set the wrong table size, bigger than actually supported, and this would just result in an exception when hpack decoder is invoked with the new header table size.

Jira: https://issues.redhat.com/browse/UNDERTOW-2255

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/proxy/ProxyHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/proxy/ProxyHandler.java
@@ -18,31 +18,6 @@
 
 package io.undertow.server.handlers.proxy;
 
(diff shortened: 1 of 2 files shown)


Kurbiel1:
how do you  use it?

fl4via:
@Kurbiel1 can you please clarify your question? How do we use what exactly? This fix just fixes an exception. Maybe you are asking how to reproduce the bug it fixes?

PR #1475, opened 2023-05-11
UNDERTOW-2228 Undertow write-timeout can cause a truncate response for request coming through keep-alive connection

TomasHofman:
https://issues.redhat.com/browse/JBEAP-24358 
https://issues.redhat.com/browse/UNDERTOW-2228

This change resets the expireTime variable in the
WriteTimeoutStreamSinkConduit at the end of each request. This should
prevent a subsequent request on the same connection to be affected by
timeout counter started during a previous request.

### DIFF
--- a/core/src/main/java/io/undertow/conduits/WriteTimeoutStreamSinkConduit.java
+++ b/core/src/main/java/io/undertow/conduits/WriteTimeoutStreamSinkConduit.java
@@ -225,6 +225,12 @@ public final class WriteTimeoutStreamSinkConduit extends AbstractStreamSinkCondu
     @Override
     public void suspendWrites() {
         super.suspendWrites();
+
+        // Reset the expireTime - we don't want to keep counting if the writes has been suspended. The expireTime will be set
+        // again during next write. This is applies for instance for remoting connections which are effectively single HTTP
+        // request, but server is suspending and resuming writes during subsequent remoting calls.
+        expireTime = -1;
+
         XnioExecutor.Key handle = this.handle;
         if(handle != null) {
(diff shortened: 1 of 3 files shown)


TomasHofman:
FYI @fl4via

fl4via (changes requested):
We will need to keep the old constructor as this is public API

TomasHofman:
Thanks! I added the original constructor back.

fl4via (approved):

TomasHofman:
I can reproduce the Windows test freeze on a Windows VM easily, but so far didn't manage to do anything about it.

Attaching stacktrace screenshot of where the test case gets stuck (didn't manage to get clipboard sharing working with the Windows VM :) ). At first sight it looks like the httpclient gets stuck when reading the response entity, but it could be caused by some discrepancy on undertow side I guess...

Screenshot from 2023-06-08 15-11-35

TomasHofman:
Hello @fl4via , the stuck Windows tests are gone. There is one failure in DefaultServletCachingListenerTestCase in the "Undertow CI / JDK 17 - servlet - ubuntu-latest" which looks unrelated to me. Locally is passing.

fl4via:
Thanks @TomasHofman !
The failure is unrelated indeed. I was able to reproduce that issue in my computer once and verify why is the file failing to refresh. This issue arises when the OS has too many open files at the same time and it can't update it. You need to change a configuration in the Linux where you say how many open files are supported. When you increase the number in that config, the failure is sorted out. And it happens intermittently because it depends on the state of the OS at the moment it runs CI.

PR #1487, opened 2023-06-07
[UNDERTOW-2010] Add invalidate() to caching manager

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2010


core/src/main/java/io/undertow/server/handlers/cache/LRUCache.java
```diff
@@ -172,6 +172,18 @@ public void clear() {
         accessQueue.clear();
     }
 
+    public Set<K> keys(){
```
ropalka, line 175: There's a keySet() method defined already serving the same purpose.
Further more it properly makes returned keys read only. 
We should remove this keys() method.

core/src/main/java/io/undertow/server/handlers/resource/CachingResourceManager.java
```diff
@@ -162,10 +185,39 @@ public void removeResourceChangeListener(ResourceChangeListener listener) {
     }
 
     public void invalidate(String path) {
+        //NOTE: dont need write here?
+        final Lock readLock = this.cacheAccessLock.readLock();
```
ropalka, line 189: Write lock should be used instead of read lock IMHO

core/src/main/java/io/undertow/server/handlers/resource/CachingResourceManager.java
```diff
@@ -162,10 +185,39 @@ public void removeResourceChangeListener(ResourceChangeListener listener) {
...
+    public void invalidate() {
+        final Lock writeLock = this.cacheAccessLock.writeLock();
+        final LRUCache<String, Object> localCopy = this.cache;
+        writeLock.lock();
+        try {
+            this.cache = new LRUCache(localCopy.getMaxEntries(),localCopy.getMaxAge());
+        } finally {
+            writeLock.unlock();
+        }
+        for(String key:localCopy.keys()) {
```
ropalka, line 207: Use localCopy.keySet() instead

core/src/main/java/io/undertow/server/handlers/resource/CachingResourceManager.java
```diff
@@ -107,43 +112,61 @@ public CachedResource getResource(final String p) throws IOException {
...
-            if(nextCheck > 0) {
-                long time = System.currentTimeMillis();
-                if(time > nextCheck) {
-                    marker.setNextCheckTime(time + maxAge);
-                    if(underlyingResourceManager.getResource(path) != null) {
-                        cache.remove(path);
+        //ReadLock will allow multiple reads and changes, guarded by WriteLock for purge
+        final Lock readLock = this.cacheAccessLock.readLock();
+        try {
+            readLock.lock();
```
ropalka, line 118: getResource() method is incorrect I would say.
It should use both read lock & write lock separately (and not just read lock).

The first part under readlock should be:
---
Object res = null
readLock.lock();
try {
    res = cache.get(path);
} finally {
    readLock.unlock();
}

And later whenever an attempt to update cache field is made, write lock should be used instead.

ropalka (changes requested):

ropalka (approved):
Good job @baranowb !

fl4via (approved):

PR #1495, opened 2023-06-22
Reproducer for request failures present in 2.2.25 and 2.3.7.

carterkozak:
I attempted a simpler reproducer without servlets and jersey, however I was unable to reproduce the issue. Similarly, when I read the response without parsing JSON, I failed to reproduce the issue. Parsing response JSON may be adding some backpressure necessary to trigger the failure.

I've shared this reproducer as a draft PR into the undertow repo for ease of use in debugging, I don't expect it to be merged.

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -39,6 +39,35 @@
             <groupId>io.undertow</groupId>
             <artifactId>undertow-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.undertow</groupId>
+            <artifactId>undertow-servlet</artifactId>
+        </dependency>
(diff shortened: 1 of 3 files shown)


baranowb:
@carterkozak  - is there a ticket for those failures?

carterkozak:
@baranowb I don't think I filed one -- I chatted with @fl4via on the Wildfly Zulipchat, though.

fl4via:
After a big void dealing with CVEs and releases, we are finally back to this one. I'll be away during next week, but then I'll pick up from here in early July.

fl4via:
I'm closing this as the issue is solved by PR #1648

PR #1506, opened 2023-07-25
[UNDERTOW-2295] Stop registration of managed default servlet in case default servlet exists

dreis2211:
Hi,

this addresses https://issues.redhat.com/projects/UNDERTOW/issues/UNDERTOW-2295, which came up while trying to mitigate https://spring.io/security/cve-2023-34035 and https://github.com/spring-projects/spring-security/issues/13568

Tests were green locally. Feel free to discuss the implications of this either here or in the ticket. This is more of a starting point to get the ball rolling. Unfortunately, I couldn't find a ticket that references the decision behind the default servlet registration, but only the commit https://github.com/undertow-io/undertow/commit/71c5a691 which is unfortunately not providing more info. 

There is https://issues.redhat.com/browse/UNDERTOW-149 which introduced a check for the managed default servlet not being registered twice, which the PR still kept.

Since this is likely a breaking change of some sort (is it?), I'm also happy to discuss configuration options.

Cheers,
Christoph

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
+++ b/servlet/src/main/java/io/undertow/servlet/handlers/ServletPathMatches.java
@@ -305,15 +305,16 @@ public class ServletPathMatches {
                 }
             }
         }
-        ServletHandler managedDefaultServlet = servlets.getServletHandler(DEFAULT_SERVLET_NAME);
-        if(managedDefaultServlet == null) {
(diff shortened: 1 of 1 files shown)


baranowb:
Looks good.

fl4via:
This changes the behavior for a handler, hence, I'm adding it temporarily to 2.5.0 payload, until we are clear on how to handle changes that sneak in the server automatically

PR #1509, opened 2023-07-31
[UNDERTOW-2298] Avoid NPE on concurrent close/timeout

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2298

Tenative PR just so we can discuss


fl4via:
@baranowb I think we should catch the NPE and double check for FLAG_FINISHED.

fl4via:
This relates to UNDERTOW-2403 #1602 .

PR #1516, opened 2023-09-25
[UNDERTOW-2312] multibytes language in URL request to http/https are …

baranowb:
…broken in EAP access log.
ISsue: https://issues.redhat.com/browse/UNDERTOW-2312

I need to check if there are TCs for access log. Im pushing this just for review ATM.

### DIFF
--- a/core/src/main/java/io/undertow/protocols/http2/HpackDecoder.java
+++ b/core/src/main/java/io/undertow/protocols/http2/HpackDecoder.java
@@ -248,7 +248,7 @@ public class HpackDecoder {
             return readHuffmanString(length, buffer);
         }
         for (int i = 0; i < length; ++i) {
-            stringBuilder.append((char) buffer.get());
+            stringBuilder.append((char) (buffer.get() & 0xFF));
         }
         String ret = stringBuilder.toString();
         stringBuilder.setLength(0);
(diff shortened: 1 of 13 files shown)


fl4via:
I'll update this PR myself as I have a test that is ready for HTTP and now I'm trying to come up with a https one.

fl4via:
Right now I'm working on making the test work with Http2 upgrade. It seems I solved the AJP test case, I'll let you know when I'm done, and I'll update this PR with my fixes.

fl4via:
I have just updated the PR with my latest fixes. I can see that proxy http2 scenario is not working, and the reason for this is that the code goes through a different path when serving the request, and it ends up not invoking HttpRequestParser. It might be an issue in the tests or in the code, at this point I am not sure, and I will need to investigate further to find out.

baranowb:
AFAIR often tests do some shady stuff to mimic "expected behavior".

baranowb:
Sometimes tests mimic desired behavior that is expected, because setup is not attainable (it seems)

fl4via:
the failure that are left are not reproducible in my machine. They will require an investigation in github ci itself

fl4via (approved):

fl4via:
The problem in Windows CI is fixed. But now we have a problem in the new LotsOfQueryParamtersTestCase test. Investigating.

jasondlee (approved):
These changes look good to me. Side note: I like the addition of `final`. :)

PR #1536, opened 2023-11-23
[UNDERTOW-1497] add request body ready bytes attribute

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1497

### DIFF
--- a/core/src/main/java/io/undertow/attribute/BytesReadAttribute.java
+++ b/core/src/main/java/io/undertow/attribute/BytesReadAttribute.java
@@ -1,0 +1,84 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -768,6 +781,10 @@ void updateBytesSent(long bytes) {
         }
     }
 
+    void updateBytesRead(long bytes) {
+        requestBytesRead+=bytes;
```
rhusar, line 785: Formatting.

core/src/main/java/io/undertow/attribute/BytesReadAttribute.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
rhusar, line 3: We can't apply copyright retroactively.

rhusar (changes requested):

core/src/main/java/io/undertow/attribute/BytesReadAttribute.java
ropalka: This was fixed with latest update.

core/src/main/java/io/undertow/server/HttpServerExchange.java
ropalka: This was fixed with latest update.

ropalka (approved):

PR #1546, opened 2024-02-06
[UNDERTOW-2338] Use a lock when setting the AsyncContext and asyncSta…

jamezp:
…rted variables. There is a race condition on when the isAsyncStarted() may return true before the asyncContext is set.

https://issues.redhat.com/browse/UNDERTOW-2338


fl4via:
@jamezp as we discussed in Zulip, I investigated if I think it can be fixed without the lock and it can. I opened a new PR: #1546  and I am marking this one as duplicate for now. If the error persists in RestEasy CI, please let me know

PR #1560, opened 2024-02-22
[UNDERTOW-1758] make state field atomic in HttpServerExchange

burov4j:
https://issues.redhat.com/browse/UNDERTOW-1758

### DIFF
--- a/core/src/main/java/io/undertow/server/HttpServerExchange.java
+++ b/core/src/main/java/io/undertow/server/HttpServerExchange.java
@@ -78,6 +78,7 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.xnio.Bits.allAreSet;
 import static org.xnio.Bits.anyAreClear;
@@ -165,7 +166,7 @@ public final class HttpServerExchange extends AbstractAttachable {
 
     // mutable state
 
-    private int state = 200;
+    private AtomicInteger state = new AtomicInteger(200);
     private HttpString requestMethod = HttpString.EMPTY;
     private String requestScheme;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -165,7 +166,7 @@ public final class HttpServerExchange extends AbstractAttachable {
 
     // mutable state
 
-    private int state = 200;
+    private AtomicInteger state = new AtomicInteger(200);
```
fl4via, line 169: @burov4j  Thanks for your PR!
I would like to follow the same pattern we have in XNIO. Could you instead make the field volatile and add an atomic field updater, like in this class: https://github.com/xnio/xnio/blob/3.x/api/src/main/java/org/xnio/Connection.java ?
This will have a better memory footprint than the currently proposed solution.

fl4via (changes requested):

core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -480,9 +483,9 @@ public HttpServerExchange setRequestURI(final String requestURI) {
     public HttpServerExchange setRequestURI(final String requestURI, boolean containsHost) {
         this.requestURI = requestURI;
         if (containsHost) {
-            this.state |= FLAG_URI_CONTAINS_HOST;
+            stateUpdater.accumulateAndGet(this, FLAG_URI_CONTAINS_HOST, (state, flag) -> state | flag);
```
carterkozak, line 486: Perhaps these lambdas could be extracted to static reusable instances -- sometimes lambdas like this in hot paths aren't elided as quickly as one might wish.

burov4j: Isn't this a premature optimization?

ropalka (approved):

core/src/main/java/io/undertow/server/HttpServerExchange.java
ropalka: Thanks @burov4j !

baranowb:
@burov4j 
1. Could you please change "state" in updater invocation to "currentState" ? its kind of misleading(though in the end its the same ref).
2. Could you please rebase and create PRs for main and 2.3.x?

baranowb (approved):

burov4j:
@baranowb looks like the issue was already fixed here: https://github.com/undertow-io/undertow/pull/1661

fl4via:
Indeed, this one is already solved! I just updated the Jiras and I'll close this PR.
Nonetheless, thank you for your contribution @burov4j ! It is greatly appreciated!

PR #1564, opened 2024-03-06
[UNDERTOW-2249] Change that HttpClientConnection.sendRequest on a clo...

xjusko:
...sed connection results in a ClosedChannelException instead of IOException
https://issues.redhat.com/browse/UNDERTOW-2249

### DIFF
--- a/core/src/main/java/io/undertow/client/UndertowClientMessages.java
+++ b/core/src/main/java/io/undertow/client/UndertowClientMessages.java
@@ -25,6 +25,7 @@ import org.jboss.logging.annotations.MessageBundle;
 
 import java.io.IOException;
 import java.net.URI;
+import java.nio.channels.ClosedChannelException;
 
 /**
  * starting from 1000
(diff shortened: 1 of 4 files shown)


xjusko:
Hi @fl4via , could you please review my PR? Thanks.

fl4via (dismissed):
@xjusko the ClosedChannelException is not always appropriate, as if you look at the usages of UndertowClientMessages.invalidConnectionState, it is thrown not only when the connection is closed or close is requested, but also on upgrade scenarios.
That being said, I see the point you raised and I think the ticket is valid. So, my suggestion here is to add another message just for closed connections, that message would throw the ClosedChannelException. Then, at the usage points, you would have two checks: one for upgrades, that uses the ld IOException invalidConnectionState, and the other one for close states, and this one would use your new message and throw the ClosedChannelException.

xjusko:
Thanks for the review @fl4via , it should be fixed based on your suggestion.

xjusko:
@fl4via I see that CI failed, but I don't think that my changes have any implications on the test that failed.

baranowb:
@xjusko ^^

ropalka (approved):

PR #1574, opened 2024-04-03
[UNDERTOW-2371] initialize the DefaultServer once to speed up test HttpContinueSslServletTestCase

TestBoost:
There are five tests in the test class `HttpContinueSslServletTestCase`. However, they reply on the same DefaultServer to send post http request. The time cost of start or stop for the DefaultServer is high. They are not modifying the contents of the DefaultServer, thus we can just initialize the DefaultServer once before or after all test methods inside test class `HttpContinueSslServletTestCase`.

The test runtime can jump from `5.43160 s` to `1.8027 s` after applying these changes when run on our machine.

### DIFF
--- a/servlet/src/test/java/io/undertow/servlet/test/handlers/AbstractHttpContinueServletTestCase.java
+++ b/servlet/src/test/java/io/undertow/servlet/test/handlers/AbstractHttpContinueServletTestCase.java
@@ -32,7 +32,6 @@ import org.apache.http.params.BasicHttpParams;
 import org.apache.http.params.HttpParams;
 import org.junit.Assert;
 import org.junit.Assume;
-import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


baranowb:
@TestBoost Could you please create a ticket?

TestBoost:
Sure, I just submit a ticket here https://issues.redhat.com/browse/UNDERTOW-2371. I don't know why the test class `Http2ClientTestCase` fails because the DefaultServer is already closed. Maybe we also need some checks in the test `Http2ClientTestCase` too.

baranowb:
@TestBoost could you please force push again to trigger CI?

ropalka (approved):

servlet/src/test/java/io/undertow/servlet/test/handlers/AbstractHttpContinueServletTestCase.java
```diff
@@ -56,10 +55,10 @@ public abstract class AbstractHttpContinueServletTestCase {
     public static void setup() {
         DeploymentUtils.setupServlet(Servlets.servlet(ContinueConsumeServlet.class).addMappings("/path"),
                 Servlets.servlet(ContinueIgnoreServlet.class).addMappings("/ignore"));
+        Assume.assumeFalse(DefaultServer.isAjp());
     }
 
-    @Before
-    public void before() throws Exception {
+    public static void before() throws Exception {
```
baranowb, line 61: Is this method even used now?

baranowb (changes requested):
If marked method is not used, it should be removed?

PR #1576, opened 2024-04-08
[UNDERTOW-1889] switch to mockito

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-1889


core/pom.xml
```diff
@@ -128,8 +128,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.easymock</groupId>
-            <artifactId>easymock</artifactId>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
```
rhusar, line 132: This makes no sense - don't add a dependency that is never used. If a subsequent change will require this dependency, the onus will be on them.

rhusar (changes requested):

core/pom.xml
ropalka: I agree with @rhusar - the org.easymock dependency should be removed if it is not used at all.

ropalka (changes requested):

fl4via:
I agree that we need to have dependencies that are only being used.
My two cents, @baranowb : if you plan to add a mock of a final class, add it to the same PR. If not, close this PR for now and reopen it when you are ready to submit a testcase that mocks a final class.

core/pom.xml
baranowb: "This actually made test for some changes, ie. [UNDERTOW-1801](https://issues.redhat.com/browse/UNDERTOW-1801) not possible."
Ok, closing.

baranowb:
easy mock isnt used AFAIR, content of ticket:
"This actually made test for some changes, ie. [UNDERTOW-1801](https://issues.redhat.com/browse/UNDERTOW-1801) not possible."

PR #1577, opened 2024-04-09
[UNDERTOW-2039] Convert AbstractFramedChannel.AWAIT_WRITABLE_TIMEOUT to nanoseconds

xjusko:
https://issues.redhat.com/browse/UNDERTOW-2039

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
@@ -62,9 +62,9 @@ public abstract class AbstractFramedStreamSinkChannel<C extends AbstractFramedCh
 
 
     /**
-     * The maximum timeout to wait on awaitWritable in milliseconds when not specified.
+     * The maximum timeout to wait on awaitWritable in nanoseconds when not specified.
      */
-    private static final int AWAIT_WRITABLE_TIMEOUT;
+    private static final long AWAIT_WRITABLE_TIMEOUT;
 
     /**
      * Extra timeout to make sure the flush has actually timed out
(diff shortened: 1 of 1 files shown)


xjusko:
Hi @fl4via , could you look at this PR if it is correctly implemented?
I am not sure about changing other timeout-related values to nanoseconds as well. 
Also, I kept the logger message in milliseconds for better readability, but let me know if I should change that to nanoseconds as well.

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
```diff
@@ -73,8 +73,8 @@ public abstract class AbstractFramedStreamSinkChannel<C extends AbstractFramedCh
 
 
     static {
-        final int defaultAwaitWritableTimeout = 600000;
-        int await_writable_timeout = AccessController.doPrivileged((PrivilegedAction<Integer>) () -> Integer.getInteger("io.undertow.await_writable_timeout", defaultAwaitWritableTimeout));
+        final long defaultAwaitWritableTimeout = 600000000000L;
```
rhusar, line 76: This is not readable, lets create sustainable software, e.g. replace with:

    final long defaultAwaitWritableTimeout = TimeUnit.MINUTES.toNanos(10);

rhusar (changes requested):

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
xjusko: You are right, I fixed it. 
Also, I am unsure about reading `io.undertow.await_writable_timeout` property in nanoseconds. I would change it to seconds or minutes as well, and then parse it to nanoseconds. But I can't find where this property can be set.

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
```diff
@@ -73,8 +73,8 @@ public abstract class AbstractFramedStreamSinkChannel<C extends AbstractFramedCh
 
 
     static {
-        final int defaultAwaitWritableTimeout = 600000;
-        int await_writable_timeout = AccessController.doPrivileged((PrivilegedAction<Integer>) () -> Integer.getInteger("io.undertow.await_writable_timeout", defaultAwaitWritableTimeout));
+        final long defaultAwaitWritableTimeout = TimeUnit.MINUTES.toNanos(10);
+        long await_writable_timeout = AccessController.doPrivileged((PrivilegedAction<Long>) () -> Long.getLong("io.undertow.await_writable_timeout", defaultAwaitWritableTimeout));
```
rhusar, line 77: This would break compatibility - users were setting `io.undertow.await_writable_timeout` thinking they are setting milis, but now that value would be taken directly and treated to nanoseconds.

rhusar (changes requested):

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedStreamSinkChannel.java
xjusko: Thanks for the answer. I was thinking that maybe I could edit some documentation to mention the change of property for better readability to seconds or minutes. But for now, we can keep the system property in milliseconds and convert it to nanos after reading it.

ropalka (approved):

rhusar:
@ropalka would be good idea to run CI as well ;-) you need to approve running.

baranowb (approved):
LGTM

PR #1578, opened 2024-04-10
[UNDERTOW-2361] handle inflater wrapping properly in deflate encoding

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2361

### DIFF
--- a/core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
+++ b/core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
@@ -32,6 +32,7 @@ import org.xnio.conduits.AbstractStreamSourceConduit;
 import org.xnio.conduits.ConduitReadableByteChannel;
 import org.xnio.conduits.StreamSourceConduit;
 import io.undertow.UndertowLogger;
+import io.undertow.UndertowMessages;
 import io.undertow.connector.PooledByteBuffer;
 import io.undertow.server.ConduitWrapper;
 import io.undertow.server.HttpServerExchange;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
```diff
@@ -72,18 +74,52 @@ public InflatingStreamSourceConduit(
             ObjectPool<Inflater> inflaterPool) {
         super(next);
```
ropalka, line 75: How about using this( ... ) constructor as above?

core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
```diff
@@ -171,6 +209,35 @@ public int read(ByteBuffer dst) throws IOException {
...
+        //            1      18 19   18 57   18 95   18 D3
+        //            2      28 15   28 53   28 91   28 CF
+        //            3      38 11   38 4F   38 8D   38 CB
+        //            4      48 0D   48 4B   48 89   48 C7
+        //            5      58 09   58 47   58 85   58 C3
+        //            6      68 05   68 43   68 81   68 DE
+        //            7      78 01   78 5E   78 9C   78 DA
+        buf.mark();
+        final char cmf = (char)(buf.get() & 0xFF);
+        final char flg = (char)(buf.get() & 0xFF);
```
ropalka, line 236: What if buf size is 1? We couldn't read 2nd byte from it then.

ropalka (changes requested):

core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
baranowb: We could, but I think JDK would not like that: "The constructor InflatingStreamSourceConduit(StreamSourceConduit) is undefined"

ropalka (approved):

flozano:
hi @fl4via , any idea when this would be released ?

flozano:
hi again @fl4via , any hint when this would be released? seems the review was approved but never merged.

fl4via:
Hi @flozano . Certainly in the next release

flozano:
this didn't make it to 2.3.18.Final ? :(

fl4via:
@flozano this will make it to the next release

flozano:
thank you @fl4via

fl4via:
Hello @flozano . This was added to 2.3.19.Final and it seems it caused a regression as you can see in #1796

PR #1585, opened 2024-04-23
[UNDERTOW-2378] Adjust properly session timeout also for alternative AUTH mechanisms

ropalka:
https://issues.redhat.com/browse/UNDERTOW-2378

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/handlers/security/ServletFormAuthenticationMechanism.java
+++ b/servlet/src/main/java/io/undertow/servlet/handlers/security/ServletFormAuthenticationMechanism.java
@@ -18,10 +18,15 @@
 
 package io.undertow.servlet.handlers.security;
 
+import static io.undertow.security.api.SecurityNotification.EventType.AUTHENTICATED;
+import static io.undertow.util.Methods.POST;
 import static io.undertow.util.StatusCodes.OK;
 
 import io.undertow.security.api.AuthenticationMechanism;
 import io.undertow.security.api.AuthenticationMechanismFactory;
+import io.undertow.security.api.SecurityContext;
+import io.undertow.security.api.NotificationReceiver;
+import io.undertow.security.api.SecurityNotification;
 import io.undertow.security.idm.IdentityManager;
 import io.undertow.security.impl.FormAuthenticationMechanism;
 import io.undertow.server.HttpServerExchange;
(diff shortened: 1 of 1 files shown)


soul2zimate:
@fl4via Could you review this ? we need this fix in downstream

fl4via (approved):

servlet/src/main/java/io/undertow/servlet/handlers/security/ServletFormAuthenticationMechanism.java
```diff
@@ -230,33 +230,57 @@ protected void storeInitialLocation(final HttpServerExchange exchange, byte[] by
...
+        Session session;
+        if (System.getSecurityManager() == null) {
+            session = httpSession.getSession();
+        } else {
+            session = AccessController.doPrivileged(new HttpSessionImpl.UnwrapSessionAction(httpSession));
+        }
+
+        if (newSession) {
+            final int originalMaxInactiveInterval = session.getMaxInactiveInterval();
+            if (originalMaxInactiveInterval > authenticationSessionTimeout) {
```
baranowb, line 272: just a minor pedantic adjustment? super.authenticationSessionTimeout ?

baranowb: Though Im not hell bent on that, its just good practice IMHO.

servlet/src/main/java/io/undertow/servlet/handlers/security/ServletFormAuthenticationMechanism.java
```diff
@@ -195,27 +215,7 @@ protected void storeInitialLocation(final HttpServerExchange exchange, byte[] by
...
-            session = AccessController.doPrivileged(new HttpSessionImpl.UnwrapSessionAction(httpSession));
-        }
-        if (newSession) {
-            int originalMaxInactiveInterval = session.getMaxInactiveInterval();
-            if (originalMaxInactiveInterval > authenticationSessionTimeout) {
-                session.setAttribute(ORIGINAL_SESSION_TIMEOUT, session.getMaxInactiveInterval());
-                session.setMaxInactiveInterval(authenticationSessionTimeout);
-            }
-        }
+        Session session = getAndInitializeSession(exchange, true);
```
baranowb, line 218: this should check for null, since its possible to get null session in some cases?

baranowb: https://github.com/undertow-io/undertow/blob/master/servlet/src/main/java/io/undertow/servlet/spec/ServletContextImpl.java#L973

baranowb (changes requested):

servlet/src/main/java/io/undertow/servlet/handlers/security/ServletFormAuthenticationMechanism.java
fl4via: I think it is different, because the getAndInitializeSession returns null only if the second parameter, create is false? Unless I am missing something. I am merging as is, if I"m wrong, let's discuss tihs further and submit another fix @baranowb

servlet/src/main/java/io/undertow/servlet/handlers/security/ServletFormAuthenticationMechanism.java
fl4via: okay, agreed! Because this is minor, though, I am merging it as is

PR #1591, opened 2024-05-09
[UNDERTOW-2389] Use WeakHashMap in DefaultByteBufferPool to avoid leak for released threads

aogburn:
https://issues.redhat.com/browse/UNDERTOW-2389

PR #1589, opened 2024-05-03
[UNDERTOW-2385] Modify ThreadLocalCache using WeakHashMap

ddaaac:
https://issues.redhat.com/browse/UNDERTOW-2385

PR #1599, opened 2024-06-10
[UNDERTOW-2385] Use synchronized WeakHashMap in DefaultByteBufferPool

fl4via:
Jira: https://issues.redhat.com/browse/UNDERTOW-2385
Duplicated Jira: https://issues.redhat.com/browse/UNDERTOW-2389
Incorporates #1589 and #1591

### DIFF
--- a/core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
+++ b/core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
@@ -26,9 +26,9 @@ import java.lang.ref.WeakReference;
 import java.nio.ByteBuffer;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/undertow/server/DefaultByteBufferPool.java
```diff
@@ -343,7 +344,7 @@ protected void finalize() throws Throwable {
     // class can be called by a different thread than the one that initialized the data.
     private static class ThreadLocalCache {
 
-        Map<Thread, ThreadLocalData> localsByThread = new HashMap<>();
+        Map<Thread, ThreadLocalData> localsByThread = Collections.synchronizedMap(new WeakHashMap<>());
```
He-Pin, line 347: how about using a concurrent hashmap?

fl4via: Good idea @He-Pin ! I'll swap it by ConcurrentHashMap. Thanks!

fl4via: Oh, I just see it now. We don't have a WeakConcurrentHashMap unfortunately... so, we would miss the functionality for WeakReferences being automatically removed. I'll have to stick with the synchronized wrapper for now. Anyway, thanks for the feedback :-)

PR #1602, opened 2024-06-13
[UNDERTOW-2403] Fix race condition in read from buffer at ServletInpu…

fl4via:
…tStreamImpl: prevent a NPE from being thrown to invoker in case input stream is closed by another thread (which can happen typically in timeouts)

Jira: https://issues.redhat.com/browse/UNDERTOW-2403

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/spec/ServletInputStreamImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/ServletInputStreamImpl.java
@@ -186,8 +186,7 @@ public class ServletInputStreamImpl extends ServletInputStream {
         int copied = Math.min(buffer.remaining(), len);
         buffer.get(b, off, copied);
         if (!buffer.hasRemaining()) {
-            pooled.close();
-            pooled = null;
+            closePoolIfNotNull();
             if (listener != null) {
                 readIntoBufferNonBlocking();
             }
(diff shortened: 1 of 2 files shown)


servlet/src/main/java/io/undertow/servlet/spec/ServletInputStreamImpl.java
```diff
@@ -196,50 +195,83 @@ public int read(final byte[] b, final int off, final int len) throws IOException
...
-                    pooled = null;
+                    closePoolIfNotNull();
                 }
             }
         }
     }
 
+    private void closePoolIfNotNull() {
+        try {
+            if (pooled != null) {
```
baranowb, line 266: ONly thing Id consider is to have local var? unless thats calculated choice to sometimes pay price of exception, rather than constant cost of local var and assignment.

baranowb: @fl4via  ^^

fl4via: We will pay the price of the NPE anyway, but, yes, I think adding the local var is a good idea because it will make the exception less likely.
`if (pooled != null) {
   final PooledByteBuffer pooled = this.pooled; // we can still read a null value here, hence, NPE when closing
   this.pooled = null;
   pooled.close();  // marginal risk of NPE here
} `

fl4via: As a secod thought, will it make it less likely, if the time spent seting it to a variable might be the same time it takes to invoke close directly? Or is it not? I'll try to see if I can find some information on this. But the code I suggested above might have the same statistical risk of getting the NPE as the one we have without the local variable.

baranowb:
Im going to trigger CI once more. Most likely this either needs to disable on mac or expect failure. Most likely we should have some sort of OS specific annotations or utility to make this kind of exclusion easier.

AkashB23:
@fl4via @baranowb  refering to the ticket and discusion on the thread, seems like this could happen in all recent versions of undertow, but we started seeing the similar exception
```
UT005090: Unexpected failure java.lang.NullPointerException: Cannot invoke "io.undertow.connector.PooledByteBuffer.getBuffer()" because "this.pooled" is null
 at io.undertow.servlet.spec.ServletInputStreamImpl.readIntoBuffer(ServletInputStreamImpl.java:202)
 at io.undertow.servlet.spec.ServletInputStreamImpl.close(ServletInputStreamImpl.java:266)
 at io.undertow.servlet.spec.HttpServletRequestImpl.closeAndDrainRequest(HttpServletRequestImpl.java:728)
 at io.undertow.servlet.core.ServletBlockingHttpExchange.close(ServletBlockingHttpExchange.java:89)
 at io.undertow.server.HttpServerExchange.endExchange(HttpServerExchange.java:1748)
 at io.undertow.server.Connectors.executeRootHandler(Connectors.java:420)
 at io.undertow.server.HttpServerExchange$1.run(HttpServerExchange.java:900)
```
only after updating from undertow 2.3.18.Final to 2.3.20.Final as part of spring boot 3.5, was there a change in either of 2.3.19 , 2.3.20 which is causing the NPE ?  and how does this impact the request processing ?

fl4via:
@AkashB23 not that I recall.
More likely, what you are noticing is caused by this PR's bug, that was already in the code before, but that has become more likely to occur as a collateral effect of other fixes (a simple change in the time the threads take to run is enough to cause this). Anyway, let's try to get this fixed asap.
If you have a reproducer, please let me know as it could be useful!

fl4via:
This relates to UNDERTOW-2298 #1509

PR #1618, opened 2024-06-24
[UNDERTOW-2110] Allow line breaks before predicate operator and befor…

baranowb:
…e handler
Issue: https://issues.redhat.com/browse/UNDERTOW-2110

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/builder/PredicatedHandlersParser.java
+++ b/core/src/main/java/io/undertow/server/handlers/builder/PredicatedHandlersParser.java
@@ -363,6 +363,11 @@ public class PredicatedHandlersParser {
                 if(token.getToken().equals(";") && tokens.peek()!=null && tokens.peek().getToken().equals(ELSE)) {
                     // something() -> predicate; ELSE predicate; - dont end processing since its followed by ELSE and its singular block
                     continue;
+                } else if (token.getToken().equals("\n") && tokens.peek() != null && isOperator(tokens.peek().getToken())) {
+                    // predicate
+                    // OPERATOR predicate
+                    // -> handeler
+                    continue;
                 }
                 handleLineEnd(string, operatorStack, output, blocks);
             } else if (isSpecialChar(token.getToken())) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/server/handlers/builder/PredicatedHandlersParser.java
```diff
@@ -363,6 +363,11 @@ static Node parse(final String string, Deque<Token> tokens, boolean topLevel) {
                 if(token.getToken().equals(";") && tokens.peek()!=null && tokens.peek().getToken().equals(ELSE)) {
                     // something() -> predicate; ELSE predicate; - dont end processing since its followed by ELSE and its singular block
                     continue;
+                } else if (token.getToken().equals("\n") && tokens.peek() != null && isOperator(tokens.peek().getToken())) {
```
ropalka, line 366: This doesn't seem correct. What if there are multiple line breaks?

core/src/test/java/io/undertow/server/handlers/PredicatedHandlersTestCase.java
```diff
@@ -159,4 +159,86 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
...
         Assert.assertEquals("foo'bar",e.getResponseHeaders().get("test-header").getFirst());
     }
+
+    @Test
+    public void testRewriteWithLineBreaks() throws IOException {
+        DefaultServer.setRootHandler(
+                Handlers.predicates(
+
+                        PredicatedHandlersParser.parse(
+                                        "path(/skipallrules)\n and true -> done\n" +
```
ropalka, line 169: Please adjust the test to include multiple line breaks in combination with carriage returns and tabs in a row.

ropalka (changes requested):

core/src/main/java/io/undertow/server/handlers/builder/PredicatedHandlersParser.java
baranowb: not allowed. Only allowed at end of handlers section. Even in OG. 
First, tokenization turns /r into /n, secondly,  if you look at if above, it does not allow multiple breaks before ELSE( which is operator).

core/src/test/java/io/undertow/server/handlers/PredicatedHandlersTestCase.java
ropalka: Closing requirement.

core/src/main/java/io/undertow/server/handlers/builder/PredicatedHandlersParser.java
ropalka: OK

ropalka (approved):

PR #1645, opened 2024-08-12
UNDERTOW-2424: Fix the request race produced by #1495

carterkozak:
The failure was the result of disagreement over the owner of a connection, after an initial thread ends an exchange and releases the connection, but continues to operate upon it with a premature `flush()` from an illegal thread (per undertow thread model).

The HttpServerExchange uses a `WriteDispatchChannel` in order to avoid interfering with the underlying connection after the exchange is completed, however the `WriteDispatchChannel` is an outer layer, around all response-wrappers. Response wrappers themselves may still interact with the underlying connections even after the `HttpReadListener` state has been reset to allow another request to be parsed. We must wrap the lower level streams with a similar function to prevent this spillover.

In practice, the failure I've encountered and reproduced was the result of `DeflatingStreamSinkConduit` invoking `next.terminateWrites()` followed by `next.flush()`. The `terminateWrites()` invocation causes the delegate `ChunkedStreamSinkConduit` to flush within `terminateWrites` which triggers the end of the exchange. Then, the following `flush()` in `DeflatingStreamSinkConduit` applies to another request entirely. This can be reproduced quickly using the reproducer branch (#1495) with a slight code modification, adding a `sleep(100)` thusly:
```diff
if (performFlushIfRequired()) {
    state |= NEXT_SHUTDOWN;
    freeBuffer();
    next.terminateWrites();
+    try {
+        Thread.sleep(100);
+    } catch (InterruptedException e) {}
    return next.flush();
} else {
```

### DIFF
--- a/core/src/main/java/io/undertow/server/DetachableStreamSinkConduit.java
+++ b/core/src/main/java/io/undertow/server/DetachableStreamSinkConduit.java
@@ -1,0 +1,191 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2024 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package io.undertow.server;
+
+import io.undertow.UndertowMessages;
+import org.xnio.XnioIoThread;
+import org.xnio.XnioWorker;
+import org.xnio.channels.StreamSourceChannel;
+import org.xnio.conduits.StreamSinkConduit;
+import org.xnio.conduits.WriteReadyHandler;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * When this conduit is considered detached it will no longer forward calls to the delegate.
+ */
+abstract class DetachableStreamSinkConduit implements StreamSinkConduit {
+
+    private final StreamSinkConduit delegate;
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -1397,7 +1397,7 @@ public StreamSinkChannel getResponseChannel() {
             final WrapperStreamSinkConduitFactory factory = new WrapperStreamSinkConduitFactory(wrappers, responseWrapperCount, this, sinkChannel.getConduit());
             sinkChannel.setConduit(factory.create());
         } else {
-            sinkChannel.setConduit(connection.getSinkConduit(this, sinkChannel.getConduit()));
+            sinkChannel.setConduit(new ExchangeLifecycleDetachableStreamSinkConduit(connection.getSinkConduit(this, sinkChannel.getConduit()), this));
```
carterkozak, line 1400: In this case there are no wrappers, so we don't necessarily need to wrap the conduit, however I opted to include the wrapper here for symmetry/safety. Please let me know if you'd prefer for this line to be reverted!

core/src/main/java/io/undertow/server/DetachableStreamSinkConduit.java
```diff
@@ -0,0 +1,191 @@
...
+
+    DetachableStreamSinkConduit(StreamSinkConduit delegate) {
+        this.delegate = delegate;
+    }
+
+    protected abstract boolean isFinished();
+
+    @Override
+    public long transferFrom(FileChannel src, long position, long count) throws IOException {
+        if (isFinished()) {
```
jasondlee, line 48: Have you done any performance testing on repeated calls to `isFinished()`? Is there any way to memoize/cache the check? Am I optimizing prematurely? :)

carterkozak: This didn't incur a measurable cost in my benchmarks. This `isFinished` check is identical to the `isFinished` check within `HttpServerExchange.WriteDispatchChannel` which handles a similar check at a slightly different layer.
This shouldn't have much performance impact because the underlying check is `state & 2048 == 2048` where `state` is a standard non-volatile field.
I considered splitting the existing `WriteDispatchChannel` implementation so that we only need to do this check in the new conduit implementation, however that seemed riskier, and this way we do still short-circuit wrapper logic after the response has completed.

carterkozak:
Trying a more precise alternative, I think this PR handles the root cause a bit better: https://github.com/undertow-io/undertow/pull/1648

carterkozak:
Closing this in favor of #1648

PR #1646, opened 2024-08-12
[UNDERTOW-2425] At ServletOutputStreamImpl synchronized workflow (lis…

aogburn:
…tener = null), prevent the buffer.flip() from not being cleared after an error during attempts to write. Also, at ServletPrintWriter, verify if no progress is being made when attempting to encode returns overflow after flushing, and mark error even if there are remaining bytes in the buffer.

Issue: https://issues.redhat.com/browse/UNDERTOW-2425

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/spec/ServletOutputStreamImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/ServletOutputStreamImpl.java
@@ -178,34 +178,14 @@ public class ServletOutputStreamImpl extends ServletOutputStream implements Buff
             int rem = buffer.remaining();
             buffer.put(b, bytesWritten + off, rem);
             buffer.flip();
-            bytesWritten += rem;
-            int bufferCount = 1;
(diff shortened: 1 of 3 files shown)


servlet/src/main/java/io/undertow/servlet/spec/ServletOutputStreamImpl.java
```diff
@@ -332,31 +352,35 @@ public void write(ByteBuffer[] buffers) throws IOException {
                     Buffers.copy(buffer, buffers, 0, buffers.length);
                 } else {
                     final ByteBuffer[] bufs = new ByteBuffer[buffers.length + 1];
-                    buffer.flip();
```
fl4via, line 335: Buffer flip should not be removed. I'll update the PR since we will be doing the release today.

fl4via (changes requested):

fl4via:
The release had to be postponed to today because of issues in UNDERTOW-2312. I'm looking how to fix the spotbugs report issues on this one, as this PR also blocks our release.

PR #1647, opened 2024-08-12
[UNDERTOW-2425] At ServletOutputStreamImpl synchronized workflow (lis…

aogburn:
…tener = null), prevent the buffer.flip() from not being cleared after an error during attempts to write. Also, at ServletPrintWriter, verify if no progress is being made when attempting to encode returns overflow after flushing, and mark error even if there are remaining bytes in the buffer.

Issue: https://issues.redhat.com/browse/UNDERTOW-2425
Main: https://github.com/undertow-io/undertow/pull/1646


fl4via:
I will be closing this PR as it is faster to create a bulk backport

PR #1660, opened 2024-09-03
[UNDERTOW-2404] Add default sorting by type and name in directory lis…

xjusko:
…ting view; enable clickable name and size column headers for custom sorting
https://issues.redhat.com/browse/UNDERTOW-2404

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/resource/DirectoryUtils.java
+++ b/core/src/main/java/io/undertow/server/handlers/resource/DirectoryUtils.java
@@ -39,8 +39,14 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.text.SimpleDateFormat;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.FormatStyle;
+import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/undertow/server/handlers/resource/DirectoryUtils.java
```diff
@@ -154,33 +183,89 @@ public static StringBuilder renderDirectoryListing(final HttpServerExchange exch
...
+                    .append(format.format((entry.getLastModified() == null) ? new Date(0L) : entry.getLastModified()))
                     .append("</td><td>--</td></tr>\n");
         }
+    }
+
+    private static Comparator<Resource> getComparator(String sortColumn, String currentSortOrder) {
+        Comparator<Resource> comparator;
+        if ("lastModified".equals(sortColumn)) {
+            comparator = Comparator.comparing(
+                    entry -> (entry.getLastModified() == null) ? new Date(0L) : entry.getLastModified()
```
baranowb, line 215: How this can be null?

baranowb: if its somehow possible, maybe just use creation date rather than 1970 ?

baranowb (changes requested):

core/src/main/java/io/undertow/server/handlers/resource/DirectoryUtils.java
xjusko: There is no creation date, but I think using a string like `Unknown` or `Not Available`. What do you think?

fl4via: It makes sense to me. As we sure that it is possible to get a null here? Maybe we even want to have something like: "-"
Such as: last modified: -

fl4via (changes requested):

fl4via (approved):

PR #1677, opened 2024-10-04
Undertow 2462

ChandanaRao13:


core/src/main/java/io/undertow/UndertowOptions.java
```diff
@@ -112,13 +112,18 @@ public class UndertowOptions {
      */
     public static final Option<Integer> MAX_COOKIES = Option.simple(UndertowOptions.class, "MAX_COOKIES", Integer.class);
 
+    /**
+     * Default value of {@link #ALLOW_EQUALS_IN_COOKIE_VALUE} option.
```
fl4via, line 116: DEFAULT_ALLOW_ENCODED_SLASH. Please update the comment

fl4via (changes requested):
Excellent! Please add the issue number to the the commit messages like this [UNDERTOW-2462],and there is a comment that needs to be updated in the Javadoc.

fl4via:
@ChandanaRao13 to quickly edit the commit messages of the last two commits do
``
git rebase -i HEAD~2
``
Then mark both commits with e as in edit and save the editor, just like you do when saving the message of a PR. Just replace the word "pick" by "e". You will see the instructions of what to do in the editor itself . To avoid ruining your branch it is a good idea to first create a backup:
``
git checkout -b UNDERTOW-2462_backup
git checkout UNDERTOW-2462
``
That way, if you make a mistake when rebasing, you can restore your branch to the previous state at any time:
``
git reset --hard upstream/main (discards everything in your branch after main)
git rebase UNDERTOW-2462_backup (restores your branch to exactly what it was)
``
Thank you for your commit!

core/src/main/java/io/undertow/UndertowOptions.java
ChandanaRao13: Updated!

ropalka (approved):

PR #1683, opened 2024-10-15
[UNDERTOW-2448] At ServletPrintWriter.write(CharBuffer) do not mark e…

fl4via:
…rror if buffer.remaining() == remaining after flush. We need to update remaining to do a proper check here.

The aforementioned == check can return true after a successfull flush, because "remaining" value is set to be the size of the bytes remaining in the buffer before encoding. If, at that stage, the buffer is empty, "remaining" value is set to buffer.capacity(). When encoding many bytes, we will end up with a full buffer. As a result, we will try to flush, and it is often the case that the buffer will be entirely cleared after flushing. This results in buffer.remaining() being the overall buffer size, i.e., buffer.capacity(). As a result, buffer.remaining() == remaining check returns true, and write is aborted, causing the broken responses we are seeing.

Jira: https://issues.redhat.com/browse/UNDERTOW-2448


jasondlee (approved):

bca111:
Hi @fl4via ,

Do you know when will this fix released?

PR #1534, opened 2023-11-07
[UNDERTOW-2333] Undertow read/write timeout should not apply to WebSockets or SSE

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2333
main: https://github.com/undertow-io/undertow/pull/1534
2.3.x: https://github.com/undertow-io/undertow/pull/1684
2.2.x: https://github.com/undertow-io/undertow/pull/1685

PR #1685, opened 2024-10-16
[UNDERTOW-2333_2.2.x] Undertow read/write timeout should not apply to WebSockets or SSE

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2333
main: https://github.com/undertow-io/undertow/pull/1534
2.3.x: https://github.com/undertow-io/undertow/pull/1684
2.2.x: https://github.com/undertow-io/undertow/pull/1685


core/src/main/java/io/undertow/websockets/core/WebSocketChannel.java
```diff
@@ -50,7 +53,8 @@
  * @author Stuart Douglas
  */
 public abstract class WebSocketChannel extends AbstractFramedChannel<WebSocketChannel, StreamSourceFrameChannel, StreamSinkFrameChannel> {
-
+    public static final String WEB_SOCKETS_READ_TIMEOUT = "io.undertow.websockets.core.WEB_SOCKETS_READ_TIMEOUT";
```
fl4via, line 56: The majority of our properties are standardized in a way that we don't have upper case letters and underscore in their names.
Also, we are not mapping their names one-to-one with packages
Cana you please replace it by something like:"io.undertow.websockets.core.read-timeout". Maybe we can even subtract the core from the property name:
"io.undertow.websockets.read-timeout"

core/src/main/java/io/undertow/websockets/core/WebSocketChannel.java
```diff
@@ -50,7 +53,8 @@
...
+    public static final String WEB_SOCKETS_WRITE_TIMEOUT = "io.undertow.websockets.core.WEB_SOCKETS_WRITE_TIMEOUT";
```
fl4via, line 57: The same applies to the name of this property

fl4via (changes requested):

fl4via (approved):

PR #1694, opened 2024-10-20
[UNDERTOW-2518] At WebSocketTimeoutTestCase, use a longer watch timeout

fl4via:
Jira:
https://issues.redhat.com/browse/UNDERTOW-2518


fl4via:
Testing if increasing the watch timeout makes the error goes away

PR #1698, opened 2024-10-23
[UNDERTOW-2519] At ProxyHandler append the non-decoded query string to the request

fl4via:
Jira:
https://issues.redhat.com/browse/UNDERTOW-2519

### DIFF
--- a/core/src/main/java/io/undertow/server/HttpServerExchange.java
+++ b/core/src/main/java/io/undertow/server/HttpServerExchange.java
@@ -211,6 +211,10 @@ public final class HttpServerExchange extends AbstractAttachable {
      * the query string
      */
     private String queryString = "";
+    /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/undertow/server/HttpServerExchange.java
```diff
@@ -590,8 +590,8 @@ public HttpServerExchange setQueryString(final String queryString) {
         // Clean leading ?
         if( queryString.length() > 0 && queryString.charAt(0) == '?' ) {
                 this.queryString = queryString.substring(1);
-            } else {
-                this.queryString = queryString;
+        } else {
+            this.queryString = queryString;
             }
```
bstansberry, line 595: Nit: it looks like this indentation should change as well.

ropalka (approved):

PR #1706, opened 2024-11-18
[UNDERTOW-2526] Deprecate DeploymentInfo methods dealing with async context timeout containing typo and introduce new alternatives without typo

ropalka:
https://issues.redhat.com/browse/UNDERTOW-2526


baranowb (approved):

baranowb:
I wonder how should those be handled in lower branches. Might look ugly with two similar methods, but breaking binary compat is a bad thing.

ropalka:
This change wasn't backported yet to any branch @baranowb so we're lucky.

jasondlee (approved):

fl4via (commented):
Thanks for pointing this out!

PR #1707, opened 2024-11-21
UNDERTOW-1881 - Add a new exchange attribute for SSL/TLS protocol version

jasondlee:
https://issues.redhat.com/browse/UNDERTOW-1881

Add and register new ExchangeAttribute implementation


baranowb:
Shouldnt this be aiming towards main?

jasondlee:
> Shouldnt this be aiming towards main?

Perhaps. I couldn't remember where they're supposed to go, but I think you're right. We target main, and @fl4via backports as needed? I'll update the target...

baranowb (commented):
As is its fine, however given comment from @msfm : https://issues.redhat.com/browse/UNDERTOW-1881?focusedId=16067599&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16067599
Im not 100% sure this will work in all cases?

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
```diff
@@ -234,7 +234,7 @@ public void run() {
      * <p>
      * DO NOT USE THIS OUTSIDE OF A TEST
      */
-    void awaitWrittenForTest() throws InterruptedException {
+    protected void awaitWrittenForTest() throws InterruptedException {
```
baranowb, line 237: Im not keen on this. Maybe just ping pong request and set special header to attribute value?
https://github.com/undertow-io/undertow/blob/main/core/src/test/java/io/undertow/server/handlers/SetAttributeTestCase.java

core/src/main/java/io/undertow/attribute/SslProtocolAttribute.java
```diff
@@ -0,0 +1,80 @@
...
+public class SslProtocolAttribute implements ExchangeAttribute {
+
+    public static final SslProtocolAttribute INSTANCE = new SslProtocolAttribute();
+
+    @Override
+    public String readAttribute(HttpServerExchange exchange) {
+        String sslProtocol = null;
+        String transportProtocol = exchange.getConnection().getTransportProtocol();
+        if ("ajp".equals(transportProtocol)) {
+            // TODO: wrong
```
baranowb, line 34: Whats the correct way?

baranowb (changes requested):

core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java
jasondlee: Can you clarify what you mean here?

baranowb: Above ^

baranowb (changes requested):

core/src/main/java/io/undertow/attribute/SslProtocolAttribute.java
jasondlee: I have included the changes from Masafumi's PR (https://github.com/undertow-io/undertow/compare/main...msfm:master_SslProtocolAttribute). The change I had (made after his initial comment on the Jira) was basically the same, but his change moves that logic into the AJP classes, where it rightly belongs.

PR #1708, opened 2024-11-29
[UNDERTOW-2532] Fix NPE when transmitting text or binary message to websocket session at the same time

raccoonback:
Hello.
I modified it in consideration of the case where NPE can occur when sending text message to websocket session.
Please check the [UNDERTOW-2532](https://issues.redhat.com/browse/UNDERTOW-2532) jira ticket for more information.
Thanks.

### DIFF
--- a/websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
+++ b/websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
@@ -244,8 +244,10 @@ final class WebSocketSessionRemoteEndpoint implements RemoteEndpoint {
         private StreamSinkFrameChannel textFrameSender;
 
         public void assertNotInFragment() {
-            if (textFrameSender != null || binaryFrameSender != null) {
-                throw JsrWebSocketMessages.MESSAGES.cannotSendInMiddleOfFragmentedMessage();
+            synchronized (this) {
+                if (textFrameSender != null || binaryFrameSender != null) {
+                    throw JsrWebSocketMessages.MESSAGES.cannotSendInMiddleOfFragmentedMessage();
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
```diff
@@ -273,50 +275,55 @@ public void sendText(final String partialMessage, final boolean isLast) throws I
...
+                }
+                if(textFrameSender == null) {
+                    textFrameSender = undertowSession.getWebSocketChannel().send(WebSocketFrameType.TEXT);
                 }
-                Channels.flushBlocking(textFrameSender);
-            } finally {
-                if (isLast) {
-                    textFrameSender = null;
+                try {
+                    Channels.writeBlocking(textFrameSender, WebSocketUtils.fromUtf8String(partialMessage));
```
ropalka, line 287: Whole section starting with Channels.writeBlocking() and ending with Channes.flushBlocking() must be executed outside of intrinsic lock section otherwise it will introduce deadlock possibility.

websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
```diff
@@ -273,50 +275,55 @@ public void sendText(final String partialMessage, final boolean isLast) throws I
...
-                    binaryFrameSender.shutdownWrites();
+
+            synchronized (this) {
+                if (textFrameSender != null) {
+                    throw JsrWebSocketMessages.MESSAGES.cannotSendInMiddleOfFragmentedMessage();
+                }
+                if (binaryFrameSender == null) {
+                    binaryFrameSender = undertowSession.getWebSocketChannel().send(WebSocketFrameType.BINARY);
+                }
+                try {
```
ropalka, line 313: Same here, see comment above.

ropalka (changes requested):

websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
raccoonback: @ropalka That's a good comment! 👍 
I will change the section so that there is no impact on the intrinsic lock.

raccoonback: @ropalka 
As you mentioned, the section `"Channels.writeBlocking() ~ Channels.flushBlocking()"` was excluded from the intrinsic lock range.

Please review again.

websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
```diff
@@ -268,57 +268,78 @@ public void sendBinary(final ByteBuffer data) throws IOException {
...
                 throw JsrWebSocketMessages.MESSAGES.messageInNull();
             }
-            if (textFrameSender != null) {
-                throw JsrWebSocketMessages.MESSAGES.cannotSendInMiddleOfFragmentedMessage();
-            }
-            if (binaryFrameSender == null) {
-                binaryFrameSender = undertowSession.getWebSocketChannel().send(WebSocketFrameType.BINARY);
-            }
+
+            StreamSinkFrameChannel sender = getBinaryFrameSender();
```
baranowb, line 298: Im going to talk about this with @ropalka . But it seems its technically possible for competing send to undercut one another, though I might be missing how this piece is being used.

websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
```diff
@@ -268,57 +268,78 @@ public void sendBinary(final ByteBuffer data) throws IOException {
...
+            finally {
                 if (isLast) {
-                    binaryFrameSender = null;
+                    clearBinaryFrameSender();
                 }
             }
             partialByte.clear();
         }
 
+        private synchronized StreamSinkFrameChannel getTextFrameSender() throws IOException {
```
baranowb, line 315: Two different senders are guarded by single instance lock. Is there a reason for such arrangement?

websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
raccoonback: I would like to share my understanding and seek your feedback.

For example, in the case of sending a **Text** message, I assume that a single message is divided into multiple frames for transmission. In this process, the `TextFrameSender` class variable is used to call `sendText(partialMessage, isLast)` multiple times, sending partial messages through the same channel. (During this process, other `send()` methods are not allowed to send messages.)

However, if partial frames are being sent from multiple threads, I believe there is a possibility that one thread might set `textFrameSender` to null first, and then another thread could call `Channels.flushBlocking(StreamSinkFrameChannel)`, leading to a NullPointerException (NPE).

To address this, I have modified the code to ensure synchronization when accessing the shared resource, the `textFrameSender` class variable.

Additionally, could you please clarify what specific scenarios you had in mind regarding `'competing sends interfering with one another'`? 
Understanding this will help me consider those cases and make further improvements accordingly.

websockets-jsr/src/main/java/io/undertow/websockets/jsr/WebSocketSessionRemoteEndpoint.java
raccoonback: Thank you for your question. 
The `textFrameSender` and `binaryFrameSender` class variables are initialized and set to `null` during partial message transmission. 
During this process, all other `send()` methods are blocked from sending messages. Specifically, while sending **partial messages** using `textFrameSender`, transmissions using `binaryFrameSender` are also **restricted**.

If we were to apply separate locks for `textFrameSender` and `binaryFrameSender`, there would be a high risk of deadlocks, especially when different methods interact with these variables concurrently. 
To mitigate this, we decided to use a single instance lock to manage both senders.

Additionally, other `send()` methods call `assertNotInFragment()` to check if a partial message is being transmitted. 
Using separate locks in this scenario could similarly lead to deadlocks. 
For these reasons, we opted for a single lock to ensure safe and consistent behavior.

raccoonback: @baranowb 
Alternatively, it seems possible to control concurrency by declaring `textFrameSender` and `binaryFrameSender` as `AtomicReference<StreamSinkFrameChannel>` without using `synchronized`.

raccoonback:
@ropalka @baranowb 
Hello,
Has there been any additional discussion on this?
I’d appreciate it if you could review it further. 😊

raccoonback:
@jasondlee @fl4via @ropalka @baranowb
Hello! 
Could you please take a look at this PR? 
The review seems to be delayed, so I would greatly appreciate your confirmation. 
Thank you!

ropalka (approved):

raccoonback:
@jasondlee @fl4via @baranowb

Hello,

I sincerely appreciate you taking the time to review my PR despite your busy schedule.

I wanted to kindly inquire about the progress of the review for this PR. If there are any additional aspects that need to be addressed, I would greatly appreciate it if you could let me know.

Thank you once again for dedicating your valuable time to this review.

raccoonback:
@jasondlee @fl4via @baranowb

Hello!
I'd like to ask when the review for this PR is expected to take place.

jasondlee (approved):

raccoonback:
@fl4via @baranowb
Hello!
The PR review is delayed, so please check!

raccoonback:
@fl4via @baranowb
Hello!
I wanted to remind you that the review for this PR has been delayed.
Please check if there are any additional changes that need to be made.

Thanks!

raccoonback:
@jasondlee @fl4via @ropalka @baranowb
Hello!
I'd like to ask about the review status and expected merge timeline for this PR. 😊

PR #1725, opened 2025-05-27
[UNDERTOW-2553] Add reuseHostHeader to ModCluster

baranowb:
Issues: https://issues.redhat.com/browse/UNDERTOW-2553
main: #1728 (and #1899; this one was created by mistake and merged by mistake, it was supposed to have gone into 2.4.x)
2.4.x: #1902

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/proxy/mod_cluster/ModCluster.java
+++ b/core/src/main/java/io/undertow/server/handlers/proxy/mod_cluster/ModCluster.java
@@ -59,6 +59,7 @@ public class ModCluster {
     private final String rankedAffinityDelimiter;
 
     private final boolean reuseXForwarded;
+    private final boolean rewriteHostHeader;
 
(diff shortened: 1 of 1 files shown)


fl4via:
Looking into this one...

baranowb:
Since we moved further, 2.3.x no longer can accept API change? so this should be done via property ?

fl4via:
> Since we moved further, 2.3.x no longer can accept API change? so this should be done via property ?

@baranowb maybe, but do we want this in 2.3.x? Do we need it? If so, I am not sure if it is done via property, because this is not UndertowOptions. It is just a builder extra parameter. I would keep it as is, but first I need to know if this is really needed in 2.3.x. I think we should keep it in 2.4.x only unless we have good reason to backport it to 2.3.x

fl4via:
We will skip this one for 2.3.x. If we ever need to revisit this decision, please reopen this PR @baranowb

PR #1727, opened 2025-06-05
[UNDERTOW-2556] fail POST expect if content length exceeds the limit

aogburn:
https://issues.redhat.com/browse/UNDERTOW-2556

2.2.x PR: #1768 
2.3.x PR: #1769

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/http/HttpContinue.java
+++ b/core/src/main/java/io/undertow/server/protocol/http/HttpContinue.java
@@ -139,9 +139,15 @@ public class HttpContinue {
             };
         }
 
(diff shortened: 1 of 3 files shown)


fl4via (approved):

fl4via:
As discussed with @aogburn there is another possibility for moving the check to an even sooner point, so I added the Waiting PR update label to indicate that this one is not ready yet.

aogburn:
Thanks @fl4via .  Made an additional change to fail earlier.  In this case, the thrown exception will be caught by https://github.com/undertow-io/undertow/blob/main/core/src/main/java/io/undertow/server/protocol/http/HttpReadListener.java#L267 for a 400.

ropalka (approved):

PR #1737, opened 2025-07-09
[UNDERTOW-2532] At WebSocketSessionRemoteEndpoint, prevent the NPE by setting a local sender field before checking for null

fl4via:
Jira: https://issues.redhat.com/browse/UNDERTOW-2532

main PR: #1736


fl4via:
Even though I created this bakcport only because the original fix, to be reverted, is already in 2.3.x, I decided that I'll create a bulk backport, in the same way I usually do for fixes that are to be entirely backported to the maintenance branches. Closing this one for now.

PR #1744, opened 2025-07-10
[UNDERTOW-2574] Make sure the allocated buffer is closed in case AbstractFramedChannel.close races with AbstractFramedChannel.receive

fl4via:
Jira: https://issues.redhat.com/browse/UNDERTOW-2574

2.2.x PR: #1768 
2.3.x PR: #1769

### DIFF
--- a/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
+++ b/core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
@@ -430,6 +430,9 @@ public abstract class AbstractFramedChannel<C extends AbstractFramedChannel<C, R
                         readData = null;
                     }
                     if(frameDataRemaining == 0) {
+                        if (isLastFrameReceived()) {
+                            forceFree = true;
+                        }
                         receiver = null;
                     }
                     return null;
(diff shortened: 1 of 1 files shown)


fl4via:
This error is happening more than 50% of the time in main JDK21 on MacOS. It has occurred on Ubuntu once but that's very rare.
The strategy here, because of the high frequency in that scenario, is to run multiple times the CI with only JDK21 MacOS core enabled. If I can no longer reproduce the race with the buffer leak error, I'll consider the bug fixed and revert the changes in CI to merge the fix.

fl4via:
Unfortunately, the error happened on the 9th attempt to run the CI. This means that the fix has made the error much less likely, in other words we hit a race condition with the new if statement, but it has not solved entirely the issue. I'm submitting a new fix on top of that with a new attempt to cover for another race involving receive and the allocated buffer vs close.

core/src/main/java/io/undertow/server/protocol/framed/AbstractFramedChannel.java
```diff
@@ -503,16 +506,21 @@ public synchronized R receive() throws IOException {
...
+                            if (forceFree || pooled.getBuffer().capacity() < 1024) {
+                                //if there is less than 1k left we don't allow it to be re-aquired
+                                readData = null;
+                            }
+                            //even though this is freed we may un-free it if we get a new packet
+                            //this prevents many small reads resulting in a large number of allocated buffers
+                            pooled.close();
+                        }
+                    } catch (IllegalStateException illegalStateException) {
+                        // ignore, it can happen if we are racing with a close that has closed the pooled
```
fl4via, line 522: verify if this can also trigger a NPE

PR #1758, opened 2025-07-15
[UNDERTOW-2370] Guard against non-servlet ctx request with ServletNam…

baranowb:
…eAttribute in access log
Issue: https://issues.redhat.com/browse/UNDERTOW-2370

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/attribute/ServletNameAttribute.java
+++ b/servlet/src/main/java/io/undertow/servlet/attribute/ServletNameAttribute.java
@@ -35,6 +35,7 @@ public class ServletNameAttribute implements ExchangeAttribute {
 
     public static final ExchangeAttribute INSTANCE = new ServletNameAttribute();
     public static final String NAME = "Servlet Name";
+    public static final String UNDEFINED = "No-Servlet";
 
     private ServletNameAttribute() {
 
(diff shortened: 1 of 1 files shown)


servlet/src/main/java/io/undertow/servlet/attribute/ServletNameAttribute.java
```diff
@@ -43,7 +44,11 @@ private ServletNameAttribute() {
     @Override
     public String readAttribute(final HttpServerExchange exchange) {
         ServletRequestContext src = exchange.getAttachment(ServletRequestContext.ATTACHMENT_KEY);
-        return src.getCurrentServlet().getManagedServlet().getServletInfo().getName();
+        if (src != null) {
+            return src.getCurrentServlet().getManagedServlet().getServletInfo().getName();
+        } else {
+            return UNDEFINED;
```
msfm, line 50: Could you explain the reasoning behind intentionally returning "No-Servlet" here? 

I was thinking that returning `null` would be sufficient, as it gets logged as `-` in the access log by default. This would also align with other ExchangeAttributes, which typically return `null` when a value isn't available.

msfm (changes requested):

ropalka (approved):

ropalka:
@fl4via please review

PR #1774, opened 2025-08-04
EAPSUP-1960: SSEHandler Header Connection is set to close.

ehsavoie:
Since the SSE connection is supposed to be long-lived and reused for sending events it should be persistent.

Issue: https://issues.redhat.com/browse/EAPSUP-1960


ehsavoie:
Superseded by https://github.com/undertow-io/undertow/pull/1775

PR #1776, opened 2025-08-27
[UNDERTOW-2593] Purge non i18n logs other than trace level

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2593

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/UndertowServletLogger.java
+++ b/servlet/src/main/java/io/undertow/servlet/UndertowServletLogger.java
@@ -140,4 +140,8 @@ public interface UndertowServletLogger extends BasicLogger {
     @Message(id = 15024, value = "Servlet %s init() method in web application %s threw exception")
     void failedToLoad(String servletName, String appName, @Cause Throwable t);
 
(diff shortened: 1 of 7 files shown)


servlet/src/main/java/io/undertow/servlet/UndertowServletLogger.java
```diff
@@ -140,4 +140,8 @@ public interface UndertowServletLogger extends BasicLogger {
     @Message(id = 15024, value = "Servlet %s init() method in web application %s threw exception")
     void failedToLoad(String servletName, String appName, @Cause Throwable t);
 
+    @LogMessage(level = ERROR)
+    @Message(id = 15025, value = "Ignoring specified SameSite mode [%s] for cookie [%s].")
+    void ignoreSameSiteCookieMode(@Cause Throwable e, String mode, String name);
```
jasondlee: Same WARN -> ERROR question.

baranowb:
changed to WARN - it was sort of protocol error, but maybe its better to treat is as runeimt error.

PR #1761, opened 2025-07-18
[UNDERTOW-2580] Fix SameSite and cookie attributes

baranowb:
Issue:  https://issues.redhat.com/browse/UNDERTOW-2580

PR #1779, opened 2025-09-05
[UNDERTOW-2580] Fix same site attribs handlinh and consolidate cookie…

baranowb:
… code

Issue: https://issues.redhat.com/browse/UNDERTOW-2580
obsolete: https://github.com/undertow-io/undertow/pull/1761  ( it had few errors as well it seems)


core/src/main/java/io/undertow/server/handlers/Cookie.java
```diff
@@ -75,6 +85,7 @@ default boolean isSameSite() {
         return false;
     }
 
+    @Deprecated(forRemoval = true)
```
baranowb: Any one understand what could be the use case to have value and set this flag to false?

baranowb:
Seems like cleanup broke it, will have to check that.

Hmm, its fine now.

PR #1789, opened 2025-09-10
[UNDERTOW-2472] Mark ENABLE_SPDY option for removal

HliasMpGH:
https://issues.redhat.com/browse/UNDERTOW-2472


fl4via:
Because this one will be backported, I'm saving it for merging after 2.4.0.Beta1 is out. That way, it will be added to 2.3.24.Final and 2.4.0.Beta2, and likely we will remove everything that is marked for removal also on Beta2.

PR #1791, opened 2025-09-10
Avoid websocket ByteBuffer copy for single buffer

schlosna:
When a websocket message is fully contained in a single `ByteBuffer` we can avoid allocating a new `ByteBuffer` and copying the contents.


baranowb:
Does this PR have a tracking ticket?

PR #1796, opened 2025-09-16
[UNDERTOW-2604]: fix request decompression when multiple buffers must be allocated

bjlaub:
I believe this fixes [UNDERTOW-2604](https://issues.redhat.com/browse/UNDERTOW-2604), and that problem was actually an issue with `InflatingStreamSourceConduit` and not a specific issue with `HttpURLConnection` as reported in the Jira issue.

This problem seems to have been introduced in https://github.com/undertow-io/undertow/pull/1578 which attempted to fix [UNDERTOW-2361](https://issues.redhat.com/browse/UNDERTOW-2361).

fix request decompression when multiple buffers must be allocated

The inflater object used to decompress a stream was possibly being reset each time a new ByteBuffer needed to be allocated, which may happen several times depending upon the request size. This causes decompression to fail for sufficiently large request body sizes.

Instead, only initialize the inflater state once per stream (until done() is called and we can safely request a new inflater from the object pool).

2.3.x PR: #1802 
2.2.x PR: #1805

### DIFF
--- a/core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
+++ b/core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
@@ -211,10 +211,14 @@ public class InflatingStreamSourceConduit extends AbstractStreamSourceConduit<St
     }
 
     protected void initializeInflater(ByteBuffer buf) throws IOException {
-        if(isZlibHeaderPresent(buf)) {
-            this.activePooledObject = this.objectPoolWrapping.allocate();
-        } else {
-            this.activePooledObject = this.objectPoolNonWrapping.allocate();
+        // ensure the activePooledObject is set only once until done() is called
+        // we don't want to reset the inflater state mid-stream
+        if (activePooledObject == null) {
+            if (isZlibHeaderPresent(buf)) {
+                this.activePooledObject = this.objectPoolWrapping.allocate();
+            } else {
+                this.activePooledObject = this.objectPoolNonWrapping.allocate();
+            }
         }
         this.inflater = this.activePooledObject.getObject();
     }
--- a/core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
+++ b/core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
@@ -73,10 +73,7 @@ public class InflatingStreamSourceConduit extends AbstractStreamSourceConduit<St
             HttpServerExchange exchange,
             StreamSourceConduit next,
             ObjectPool<Inflater> inflaterPool) {
-        super(next);
-        this.exchange = exchange;
-        this.objectPoolNonWrapping = inflaterPool;
-        this.objectPoolWrapping = null;
+        this(exchange, next, inflaterPool, newInstanceWrappingInflaterPool());
     }
 
     public InflatingStreamSourceConduit(
--- a/core/src/test/java/io/undertow/server/handlers/encoding/RequestContentEncodingTestCase.java
+++ b/core/src/test/java/io/undertow/server/handlers/encoding/RequestContentEncodingTestCase.java
@@ -18,16 +18,13 @@
 
 package io.undertow.server.handlers.encoding;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Random;
 import java.util.zip.Deflater;
-import java.util.zip.GZIPOutputStream;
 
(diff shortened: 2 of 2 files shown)


core/src/test/java/io/undertow/server/handlers/encoding/RequestContentEncodingTestCase.java
```diff
@@ -103,6 +111,14 @@ public void testDeflateEncoding() throws IOException {
...
         runTest("Hello World", "deflate");
 
+        // the default buffer size used for tests is 8kb
+        // ensure we send enough data to require multiple buffer allocations when deflating
+        Random random = new Random();
+        byte[] randomBytes = new byte[65536];
+        random.nextBytes(randomBytes);
+        Base64.Encoder encoder = Base64.getEncoder();
+        String randomString = encoder.encodeToString(randomBytes);
+        runTest(randomString, "deflate");
```
bjlaub, line 121: these tests fail on 2.3.19.Final and pass on 2.3.18.Final (and with this fix).

bjlaub:
@baranowb @fl4via mind taking a look whenever you have a chance? Thanks!

core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
```diff
@@ -211,10 +211,14 @@ public int read(ByteBuffer dst) throws IOException {
...
-        } else {
-            this.activePooledObject = this.objectPoolNonWrapping.allocate();
+        // ensure the activePooledObject is set only once until done() is called
+        // we don't want to reset the inflater state mid-stream
+        if (activePooledObject == null) {
+            if (isZlibHeaderPresent(buf)) {
+                this.activePooledObject = this.objectPoolWrapping.allocate();
+            } else {
+                this.activePooledObject = this.objectPoolNonWrapping.allocate();
+            }
```
jamezp, line 221: This should likely use some kind of lock or it could still be reset.

baranowb:
@bjlaub 

[INFO] Starting audit...
Error:  /home/runner/work/undertow/undertow/core/src/test/java/io/undertow/server/handlers/encoding/RequestContentEncodingTestCase.java:21:8: Unused import - java.io.ByteArrayOutputStream. [UnusedImports]
Error:  /home/runner/work/undertow/undertow/core/src/test/java/io/undertow/server/handlers/encoding/RequestContentEncodingTestCase.java:23:8: Unused import - java.io.OutputStream. [UnusedImports]
Error:  /home/runner/work/undertow/undertow/core/src/test/java/io/undertow/server/handlers/encoding/RequestContentEncodingTestCase.java:30:8: Unused import - java.util.zip.GZIPOutputStream.

core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
baranowb: AFAIR read will happen sequentially in IO thread, so there is no need for that? @fl4via ^^ ?

baranowb (approved):

bjlaub:
There's a separate potential null pointer exception with this class if we are initialized via this constructor: https://github.com/undertow-io/undertow/blob/65519cf315b69ac2b3499f39dcf788bc6f401bc8/core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java#L72-L80 (which sets `objectPoolWrapping` to null and it is not surrounded by null checks elsewhere).

I'll see if I can get that fixed up as well.

core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
fl4via: @baranowb is right. We are not supposed to have multiple threads invoking the same conduit at the same time, except for a concurrent close call

core/src/test/java/io/undertow/server/handlers/encoding/RequestContentEncodingTestCase.java
fl4via: great! Thanks for adding a test @bjlaub !

fl4via (approved):

core/src/main/java/io/undertow/conduits/InflatingStreamSourceConduit.java
jamezp: Sorry, I should probably not have commented that :) I just saw the variable was volatile and assumed it could be accessed by multiple threads at once. If I'm wrong, then nothing is needed at all :)

My concern was the read for the `null` check was it could be `null` then there is a race to set it, but only if multiple threads are invoking the method.

fmiguelez:
This fix also fixes UNDERTOW-2620. I can not link it or add any comment on Jira. @fl4via, can you mark it as duplicated of this one or fixed?

PR #1934, opened 2026-03-10
[UNDERTOW-2736] Align handlers$Builder pattern

baranowb:
Issue: https://issues.redhat.com/browse/UNDERTOW-2736

### DIFF
--- a/core/src/main/java/io/undertow/server/handlers/ReasonPhraseHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/ReasonPhraseHandler.java
@@ -18,9 +18,16 @@
 
 package io.undertow.server.handlers;
 
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
 import io.undertow.UndertowLogger;
+import io.undertow.server.HandlerWrapper;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
+import io.undertow.server.handlers.builder.HandlerBuilder;
 
 /**
  * A handler which simply sets a response code.
@@ -62,5 +69,42 @@ public final class ReasonPhraseHandler implements HttpHandler {
     @Override
     public String toString() {
         return "reason-phrase( " + this.reasonPhrase + " )";
+    }
+
+    public static class Builder implements HandlerBuilder {
+        @Override
+        public String name() {
+            return "reason-phrase";
+        }
+
+        @Override
+        public Map<String, Class<?>> parameters() {
+            Map<String, Class<?>> parameters = new HashMap<>();
+            parameters.put("value", String.class);
+            return parameters;
+        }
+
+        @Override
+        public Set<String> requiredParameters() {
+            final Set<String> req = new HashSet<>();
+            req.add("value");
+            return req;
+        }
+
+        @Override
+        public String defaultParameter() {
+            return "value";
+        }
+
+        @Override
+        public HandlerWrapper build(final Map<String, Object> config) {
+            final String value = (String) config.get("value");
+            return new HandlerWrapper() {
+                @Override
+                public HttpHandler wrap(HttpHandler handler) {
+                    return new ReasonPhraseHandler(handler, value);
+                }
+            };
+        }
     }
 }
--- a/core/src/main/java/io/undertow/server/handlers/ReasonPhraseHandler.java
+++ b/core/src/main/java/io/undertow/server/handlers/ReasonPhraseHandler.java
@@ -18,9 +18,14 @@
 
 package io.undertow.server.handlers;
 
+import java.util.Map;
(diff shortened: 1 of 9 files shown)


baranowb:
As is it does not remove old classes, but removes them from services def, as it cause duplicates.

ropalka (approved):

jamezp:
Are any of these in a released version of Undertow? If not, just moving the builders. If they're only in main, at least AFAIK, we don't have backwards compatibility to worry about. However, if they're in a released version of Undertow than we do need to deprecate them.

That said, we should add the `@Deprecated` annotation and we should likely delegate the deprecated build to the non-deprecated builder so fixes only need to happen in one place.

baranowb:
yep, this has been like that for long time, Ive deprecated classes , but that might not be suitable option since its JDK service?
@jamezp ^^

ropalka:
@jamezp @fl4via @jasondlee please review. We need at least 2 reviewers to agree with proposed changes.

core/src/main/java/io/undertow/server/handlers/builder/RewriteHandlerBuilder.java
```diff
@@ -18,22 +18,25 @@
...
-import java.util.Collections;
-import java.util.Map;
-import java.util.Set;
 
 /**
  * @author Stuart Douglas
+ * @see {@link RewriteHandler#Builder}
+ * @Deprecated(since = "2.5.0", forRemoval = true)
  */
 public class RewriteHandlerBuilder implements HandlerBuilder {
```
jasondlee, line 41: Missing annotation?

core/src/main/java/io/undertow/server/handlers/ReasonPhraseHandler.java
```diff
@@ -63,4 +70,41 @@ public void handleRequest(final HttpServerExchange exchange) throws Exception {
...
+            parameters.put("value", String.class);
+            return parameters;
+        }
+
+        @Override
+        public Set<String> requiredParameters() {
+            final Set<String> req = new HashSet<>();
+            req.add("value");
+            return req;
+        }
```
jasondlee, line 92: Are these Maps and Sets supposed to be immutable? Could we use Map.of() and Set.of() to simplify the code and ensure immutability?

core/src/main/java/io/undertow/server/handlers/RewriteHandler.java
```diff
@@ -0,0 +1,126 @@
...
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        UndertowLogger.PREDICATE_LOGGER.debugf("Request rewritten to [%s] for %s.", getValue().readAttribute(exchange),
+                exchange);
+        super.handleRequest(exchange);
+    }
+
+    @Override
+    public String toString() {
+        return "rewrite( '" + getValue().toString() + "' )";
```
jasondlee, line 53: Unnecessary `.toString()`

core/src/main/java/io/undertow/server/handlers/RewriteHandler.java
```diff
@@ -0,0 +1,126 @@
...
+    public static class Builder implements HandlerBuilder {
+
+        @Override
+        public String name() {
+            return "rewrite";
+        }
+
+        @Override
+        public Map<String, Class<?>> parameters() {
+            return Collections.<String, Class<?>> singletonMap("value", ExchangeAttribute.class);
```
jasondlee, line 65: This makes me think that using `Map.of()` and `Set.of()` is the appropriate and cleaner approach.

jasondlee (changes requested):
Looks good. A couple of comments/proposals.

core/src/main/java/io/undertow/server/handlers/builder/RewriteHandlerBuilder.java
ropalka: Good catch @jasondlee !

core/src/main/java/io/undertow/server/handlers/RewriteHandler.java
baranowb: well. AFAIR this will get the "toString" callout during compile time?

jamezp: There is always a potential that `getValue().toString()` could throw an NPE. I don't know in this specific case, but either way when you concatenate any object into a string an implicit `String.valueOf(instance)` will be invoked.

ropalka: I agree that .toString() is unnecessary. And it is being evaluated at run-time and not compile-time.

ropalka:
@jamezp @jasondlee please review again. AFAICS your feedback have been included in latest push.

PR #1936, opened 2026-03-13
UNDERTOW-1947: upgrade to httpcomponents 5.6, replace depracated classes

michpetrov:
Issue: [UNDERTOW-1947](https://issues.redhat.com/browse/UNDERTOW-1947)

I understand this is a large PR, 5.x primarily went from
```
response = client.execute(get)
// do something with response
```
to
```
client.execute(get, response -> {
    // do something with response
})
```
which unfortunately means a lot of changed code. As you will see most of the response handling is basically checking the response code and maybe checking the response body, so we could expand the TestHttpClient to include some standard handling methods but that would make it even more complicated if I put it into this PR.


michpetrov:
@baranowb did something change about `ALLOW_UNESCAPED_CHARACTERS_IN_URL`? Some tests are failing that weren't before - I've added a method to percent-encode the URI, except when allow_unescaped was enabled - https://github.com/michpetrov/undertow/blob/5eea6f701a764d77dff8ed1e0fcaba5137e4b5d1/core/src/test/java/io/undertow/server/AbstractRedirectURIParseTest.java#L62
now `RedirectURIParseWithAllowUnescapedCharactersTestCase` fails unless I remove the override but that defeats the purpose of the test.

michpetrov:
I've double checked and the tests pass in the current branch, they start failing after I rebase on main.

michpetrov:
Blocked by #1970

PR #1938, opened 2026-03-17
[UNDERTOW-2738][UNDERTOW-2739][UNDERTOW-2740][UNDERTOW-2741] Fixes for Cookie header parsing

fl4via:
Jiras:

https://redhat.atlassian.net/browse/UNDERTOW-2738
https://redhat.atlassian.net/browse/UNDERTOW-2739
https://redhat.atlassian.net/browse/UNDERTOW-2740
https://redhat.atlassian.net/browse/UNDERTOW-2741

Main PR: #1956

### DIFF
--- a/core/src/main/java/io/undertow/util/Cookies.java
+++ b/core/src/main/java/io/undertow/util/Cookies.java
@@ -20,9 +20,11 @@ package io.undertow.util;
 
 import io.undertow.UndertowLogger;
 import io.undertow.UndertowMessages;
+import io.undertow.UndertowOptions;
 import io.undertow.server.CookieStore;
 import io.undertow.server.handlers.Cookie;
 import io.undertow.server.handlers.CookieImpl;
+import org.xnio.OptionMap;
 
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/undertow/util/Cookies.java
```diff
@@ -219,89 +221,61 @@ public static Map<String, Cookie> parseRequestCookies(int maxCookies, boolean al
...
+        else {
+            maxCookies = options.get(UndertowOptions.MAX_COOKIES, UndertowOptions.DEFAULT_MAX_COOKIES);
+            allowEqualInValue = options.get(UndertowOptions.ALLOW_EQUALS_IN_COOKIE_VALUE, false);
+            enableRfc6265CookieValidation = options.get(UndertowOptions.ENABLE_RFC6265_COOKIE_VALIDATION, UndertowOptions.DEFAULT_ENABLE_RFC6265_COOKIE_VALIDATION);
+        }
+        for (String cookie : cookies) {
+            parseCookie(cookie, requestCookieStore, maxCookies, allowEqualInValue, commaIsSeparator, allowHttpSeparatorsInV0, enableRfc6265CookieValidation);
+        }
+        return requestCookieStore.asLegacyMap();
     }
```
jamezp: Do we want these to return a `Map`? These are flat maps and aren't necessarily cheap. We call it from the `HttpServerExchange` which means the map will be created a lot. Granted it should get thrown away given we don't use the value, but it's a lot of small objects potentially being created.

core/src/main/java/io/undertow/util/Cookies.java
```diff
@@ -219,89 +221,61 @@ public static Map<String, Cookie> parseRequestCookies(int maxCookies, boolean al
 
     @Deprecated(since="2.4.0", forRemoval=true)
     public static void parseRequestCookies(int maxCookies, boolean allowEqualInValue, List<String> cookies, Set<Cookie> parsedCookies) {
-        parseRequestCookies(maxCookies, allowEqualInValue, cookies, parsedCookies, LegacyCookieSupport.COMMA_IS_SEPARATOR, LegacyCookieSupport.ALLOW_HTTP_SEPARATORS_IN_V0);
+        Map parsedCookiesMap = parseRequestCookies(maxCookies, allowEqualInValue, cookies, LegacyCookieSupport.COMMA_IS_SEPARATOR, LegacyCookieSupport.ALLOW_HTTP_SEPARATORS_IN_V0);
```
jamezp: Minor, but this should probably be something more like:
```java
final CookieStore cookieStore = new CookieStore();
parseRequestCookies(cookies, cookieStore, OptionMap.builder().set(UndertowOptions.MAX_COOKIES, maxCookies).set(UndertowOptions.ALLOW_EQUALS_IN_COOKIE_VALUE, allowEqualInValue).getMap());
cookieStore.forEach(parsedCookies::add);
```

Otherwise you end up with a flat map of cookies. Maybe that is what we want here though.

core/src/main/java/io/undertow/util/Cookies.java
fl4via: We don't. That's a very good catch! I'm updating the PR.

fl4via: Updated

core/src/main/java/io/undertow/util/Cookies.java
fl4via: Updated

jamezp (approved):

PR #1948, opened 2026-04-21
[UNDERTOW-2741] Cookies does not strip quotes from legacy attribute values

christosderetzis:
Issue: https://redhat.atlassian.net/browse/UNDERTOW-2741


fl4via:
Unfortunately we already have a fix for this one @christosderetzis : #1948 . We need to keep main consistent with 2.4.x and for that reason I'll be fetching that fix to main as well. My apologies for that.

PR #1951, opened 2026-05-08
[UNDERTOW-2761] return 413 status instead of 500 in case of RequestTooBigException

aogburn:
https://redhat.atlassian.net/browse/UNDERTOW-2761


grgrzybek (approved):
`REQUEST_ENTITY_TOO_LARGE` status for `RequestTooBigException` is fine - thanks!

aogburn:
Would we prefer changing each location a RequestTooBigException is thrown to set a desired response code instead of setting the response code in a common location where the RequestTooBigException is caught?  Any thoughts on that @fl4via ?

baranowb:
As to question - RootHandler should take care of it, unless someone handles IOException in general.

baranowb (approved):

ropalka (approved):

fl4via (approved):

PR #1947, opened 2026-04-21
[UNDERTOW-2594][UNDERTOW-2595][UNDERTOW-2596] CVE-2026-28368 CVE-2026…

fl4via:
…-28369 CVE-2026-28367 Switching to strict HTTP parser

Jiras:
https://redhat.atlassian.net/browse/UNDERTOW-2594
https://redhat.atlassian.net/browse/UNDERTOW-2595
https://redhat.atlassian.net/browse/UNDERTOW-2596

Main PR: #1949 
2.3.x PR: #1963 
2.2.x PR: #1965

PR #1952, opened 2026-05-13
[UNDERTOW-2763] As per RFC 9112 reason phrase is optional in HTTP responses

ropalka:
https://redhat.atlassian.net/browse/UNDERTOW-2763

Main PR: #1967 
2.3.x PR: todo
2.2.x PR: todo

### DIFF
--- a/core/src/main/java/io/undertow/client/http/ResponseParser.java
+++ b/core/src/main/java/io/undertow/client/http/ResponseParser.java
@@ -136,10 +136,10 @@ final class ResponseParser {
             if (state.previousByte == 0 && (isVisibleAsciiChar(nextByte) || isSpaceOrTabChar(nextByte) || isObsoleteChar(nextByte))) {
                 // reason phrase data goes on
                 sb.append(nextChar);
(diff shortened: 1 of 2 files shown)


rhusar (approved):
Brief review, but LGTM; crux of the fix is removal of `&& sb.length() > 0`.

I was going to say, this is missing split tests for RESPONSE2 and RESPONSE3 however, those are subsets of the testOneCharacterAtATime_response1/2/3, so I don't think they are needed.

rhusar:
Actually, @ropalka this is modifying a CVE fix so can you explain why relaxing this specific check is still safe as that appeard to be intentional strictness of the parser introduced in https://github.com/undertow-io/undertow/pull/1947? 

I don't actually see that PR adding tests for this specifically, so the current tests are might not be enough to verify this doesn't reintroduce the CVE.

ropalka:
The strictness of expecting reason-phrase to have lenght >= 1 was an oversight @rhusar .
The test updated in this PR introduced RESPONSE2 & RESPONSE3 - both are omitting reason phrase to ensure it is tested.

PR #1953, opened 2026-05-13
[UNDERTOW-2759] Use spotbugs maven plugin version that is compatible with JDK25

ropalka:
Jira: https://redhat.atlassian.net/browse/UNDERTOW-2759

Follow up for #1941 

2.2.x PR: todo
2.3.x PR: todo
2.4.x PR: todo


fl4via (changes requested):
@ropalka please move the PR to main and then I'll backport it with #1941 to save CI resources
I think we can use the same Jira for both PRs, UNDERTOW-2759
Notice that this PR has failed CI because there are findbugs errors that need to be sorted out or added to the exclude list.

PR #1971, opened 2026-05-20
[UNDERTOW-2693] Sync sameSite flag with cookie SameSite attribute in …

Croway:
…ServletCookieAdaptor

The `UNDERTOW-2693` fix re-introduced a sameSite boolean field in ServletCookieAdaptor but left two gaps where it gets out of sync with the wrapped cookie's SameSite attribute:

1. Constructor — when wrapping a jakarta.servlet.http.Cookie that already has setAttribute("SameSite", "Lax") set, the boolean stays false. Connectors gates Set-Cookie serialization on isSameSite(), so the attribute is silently dropped.
2. setAttribute() — calling setAttribute("SameSite", value) on the adaptor delegates to the wrapped cookie but doesn't sync the boolean, creating the same inconsistency.

This fix initializes sameSite from the wrapped cookie in the constructor and syncs it in setAttribute().

The issue was found while testing Apache Camel, in particular https://github.com/apache/camel-spring-boot/blob/main/components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpCookiesTest.java#L125 with undertow

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/spec/ServletCookieAdaptor.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/ServletCookieAdaptor.java
@@ -42,6 +42,7 @@ public class ServletCookieAdaptor implements Cookie {
 
     public ServletCookieAdaptor(final jakarta.servlet.http.Cookie cookie) {
         this.cookie = cookie;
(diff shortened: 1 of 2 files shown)


ropalka:
@jamezp please review

PR #1975, opened 2026-06-11
[UNDERTOW-1666] BufferOverflowException instead of IOException

HliasMpGH:
Issue: https://redhat.atlassian.net/browse/UNDERTOW-1666

I also added a new test class that is responsible for AJP response testing. I didn't add a test case that checks the exception is triggered through an AJP request, because (as far as I understand), these server-side exceptions are transformed to `ClosedChannelException` and thus the exception message/reason gets lost. Also, since `putHttpString()` and `putString()` methods are both private, I think this is the best way that we can test this new behaviour.


HliasMpGH:
I see some runtime buffer leaks in the new test case, I will look into those

HliasMpGH:
The 2 failing test cases are most likely flaky. @ropalka wdyt?

PR #1981, opened 2026-07-15
[UNDERTOW-2785] sync ServletOutputStream buffer operations to protect for race condition and crash potential

aogburn:
https://redhat.atlassian.net/browse/UNDERTOW-2785

### DIFF
--- a/servlet/src/main/java/io/undertow/servlet/spec/ServletOutputStreamImpl.java
+++ b/servlet/src/main/java/io/undertow/servlet/spec/ServletOutputStreamImpl.java
@@ -134,7 +134,7 @@ public class ServletOutputStreamImpl extends ServletOutputStream implements Buff
     /**
      * {@inheritDoc}
      */
-    public void write(final byte[] b, final int off, final int len) throws IOException {
+    public synchronized void write(final byte[] b, final int off, final int len) throws IOException {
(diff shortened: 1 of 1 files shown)


baranowb:
@aogburn - please have a look at spotbugs output.

baranowb:
Also, after quick glance over changes - Im not sure thats the proper way, as usually those kind of classes lock on specific var/synchronizer, rather than class instance. @fl4via  ^^

aogburn:
> Also, after quick glance over changes - Im not sure thats the proper way, as usually those kind of classes lock on specific var/synchronizer, rather than class instance. @fl4via ^^

Yeah, much of my motivation there was to start testing a quick and simple fix against my reproductions and confirm this resolves them, which it does.  And given the similarities this has to UNDERTOW-1839, I started with its fix approach taken in https://github.com/undertow-io/undertow/pull/1034, which was pretty broad synchronization on the methods/class instance.  Like UNDERTOW-1839, most use cases will likely have no actual concurrency here so it's probably most often an uncontended lock here as well. But I can do a pass at another fix approach limiting the sync more so to the key buffer use points with a specific lock var.

aogburn:
Closing per https://redhat.atlassian.net/browse/UNDERTOW-2785?focusedCommentId=18022546

